PR #17, opened 2016-06-03
Replace browserify with webpack

psiemens:


blx:
@psiemens u seriously have 3 builds set up across two CI platforms? 🙏

psiemens:
@blx ugh i was experimenting with different CIs. Circle shouldn't be there anymore

PR #71, opened 2016-09-27
Clean up and improve webpack, gulp configuration

psiemens:
- improved webpack configuration for smaller production builds
- added javascript and css source maps for development
- cleaned up gulpfile to make the build process more clear

@annygakh can you look over this code?

@mchernoff it might be worth trying to run `gulp` again with this new configuration -- maybe it will fix your problem?

Note: I changed the default gulp task to a watch task, so you should only have to run `gulp` once and it will then watch for changes to the source files and rebuild the project whenever you make an update.

### DIFF
--- a/ubyssey/static/gulpfile.js
+++ b/ubyssey/static/gulpfile.js
@@ -2,7 +2,10 @@ var gulp = require('gulp');
 var gutil = require('gulp-util');
 var clean = require('gulp-clean');
 
-var webpack = require('webpack-stream');
+var webpack = require('webpack');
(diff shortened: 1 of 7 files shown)


annygakh:
Peter! Sorry for the late reply, I did not get the notifications (or rather perhaps the notification looked like every other notification I get when you close a ticket or make changes).
Uhm its a bit too late but yes, code related to sourcemaps LGTM, the changes you made on my computer are the same.

psiemens:
No problem! Thanks @annygakh!

PR #75, opened 2016-09-29
#47 Fixed the issue with long bullet points

### LINKED ISSUE #47: Long bullet point items covered by right-hand ads

annygakh:
Fixed the issue with long bullet points being covered by ads on the right hand side. Now the bullet points have the same max width as the other elements of the article.

Fixes #47


psiemens (approved):
This is great! Tested in all browsers and works perfectly. Thanks @annygakh 👍

annygakh:
Sorry Peter, I forgot to add that I also tested in different browsers and in mobile mode! Thanks for taking a look!!

PR #86, opened 2016-10-13
Archive

mchernoff:
Archive page implementation includes one new HTML and one SCSS file. Added link to archive in home page, next to Volunteer, Contact, and About.

Based on search page, but includes filters for year and section.


ubyssey/templates/headers/main.html
```diff
@@ -35,6 +35,7 @@
                         <li><a href="{% url 'page' 'volunteer' %}">Volunteer</a></li>
                         <li><a href="{% url 'page' 'about' %}">About</a></li>
                         <li><a href="{% url 'page' 'contact' %}">Contact</a></li>
+						<li><a href="{% url 'archive' %}">Archive</a></li>
```
psiemens: Fix tab spacing here

ubyssey/templates/headers/mobile.html
```diff
@@ -23,6 +23,7 @@
             <li><a href="{% url 'page' 'volunteer' %}">Volunteer</a></li>
             <li><a href="{% url 'page' 'about' %}">About</a></li>
             <li><a href="{% url 'page' 'contact' %}">Contact</a></li>
+			<li><a href="{% url 'archive' %}">Archive</a></li>
```
psiemens: Fix tab spacing here

ubyssey/views.py
```diff
@@ -217,6 +216,55 @@ def author_articles(self, request, slug=None):
 
         return render(request, 'author/articles.html', context)
 
+    def archive(self, request):
+            
+        current_year = datetime.today().year
```
psiemens: Instead of returning every year from 2011 until now, I think it would be better to return a list of all the years that contain articles. This should be possible with a single query that selects all the unique years in the in the Article table.

This makes the code more robust and allows us to easily add articles for years before 2011 (if we decide to do that in the future).

ubyssey/views.py
```diff
@@ -217,6 +216,55 @@ def author_articles(self, request, slug=None):
...
+            'sections': sections,
+            'years': years
+        }
+            
+        query = request.GET.get('q', None)
+        section_id = request.GET.get('section_id', None)
+        
+        year = request.GET.get('year', None)
+        if year is None:
+            year = years[0]
```
psiemens: If we are combining the search and article pages, I think it makes more sense to _not_ filter by year if no year is specified in the request.

ubyssey/views.py
```diff
@@ -217,6 +216,55 @@ def author_articles(self, request, slug=None):
...
+        page = request.GET.get('page')
+        try:
+            articles = paginator.page(page)
+        except PageNotAnInteger:
+               # If page is not an integer, deliver first page.
+            articles = paginator.page(1)
+        except EmptyPage:
+            # If page is out of range (e.g. 9999), deliver last page of results.
+            articles = paginator.page(paginator.num_pages)
+        context['articles'] = articles
```
psiemens: Add an extra space here (for readability)

ubyssey/templates/archive.html
```diff
@@ -0,0 +1,62 @@
...
+{% extends}
+{% load staticfiles %}
+{% block header %}
+	{% include 'headers/topbar.html' with article=article section=article.section.slug %}
+	{% include 'headers/mobile.html' %}
+{% endblock %}
+{% block content %}
+    <main class="archive">
+        <div class="container padded">
+			<h1>Ubyssey Archive</h1>
```
psiemens: It looks like there's an extra tab here

ubyssey/static/src/styles/modules/_archive.scss
```diff
@@ -0,0 +1,27 @@
+.bars {
```
psiemens: All the styles in this file should ideally be under a single parent (i.e. `.archive`) to prevent conflicts with styles on other parts of the site.

ubyssey/static/src/styles/modules/_archive.scss
```diff
@@ -0,0 +1,27 @@
...
+	margin-top: 30px;
+}
+.bars .bar{
+	float: left;
```
psiemens: Let's use flexbox instead of floats for this layout

psiemens (commented):
Looking good! 

A few things:
- we should figure out a responsive design for phones and tablets
- combine search and archive page into single page
- CSS is a little wonky and causes the content to wrap below the footer -- I think this can be fixed by using a `flexbox` instead of `float`

ubyssey/views.py
mchernoff: Fixed, replaced with raw query

ubyssey/views.py
mchernoff: Fixed

ubyssey/templates/headers/main.html
mchernoff: Fixed

ubyssey/templates/headers/mobile.html
mchernoff: Fixed

ubyssey/views.py
mchernoff: Added

ubyssey/templates/archive.html
mchernoff: Fixed

ubyssey/static/src/styles/modules/_archive.scss
mchernoff: Fixed

ubyssey/static/src/styles/modules/_archive.scss
mchernoff: Fixed

ubyssey/views.py
```diff
@@ -217,6 +216,58 @@ def author_articles(self, request, slug=None):
...
+        page = request.GET.get('page')
+        try:
+            articles = paginator.page(page)
+        except PageNotAnInteger:
+               # If page is not an integer, deliver first page.
+            articles = paginator.page(1)
+        except EmptyPage:
+            # If page is out of range (e.g. 9999), deliver last page of results.
+            articles = paginator.page(paginator.num_pages)
+
```
psiemens: `context` should include a `meta` property that defines the page title, something like this:

``` python
context['meta'] = { 'title': 'Archive' }
```

Here's an example:
 https://github.com/ubyssey/ubyssey-dispatch-theme/blob/staging/ubyssey/views.py#L286-L288

mchernoff: Fixed

psiemens:
Changes to be made:
- [x] Keep applied filters when using search bar (use hidden form fields)
- [x] Indicate which filters are currently applied 
- [x] Add ability to remove specific filters or clear all
- [x] Add "sort by" dropdown
- [x] Show number of results for query (like on search page)
- [x] Combine search and archive pages

Issue #58, opened 2016-09-23
Add Facebook "like" button to each article


psiemens:
We want to add a simple "like" button to each article page. 

Facebook has a simple plugin you should be able to make use of https://developers.facebook.com/docs/plugins/like-button.

@seanhsiung94 Let's start by placing the button at the bottom of each article and see how it looks.

PR #88, opened 2016-10-13
Facebook like and share button added to the end of article.

seanhsiung94:
implemented #58 
Needs to be tested on production.


ubyssey/templates/base.html
```diff
@@ -43,6 +43,16 @@
...
+<div id="fb-root"></div>
+<script>(function(d, s, id) {
+    var js, fjs = d.getElementsByTagName(s)[0];
+    if (d.getElementById(id)) return;
+    js = d.createElement(s); js.id = id;
+    js.src = "//connect.facebook.net/en_US/sdk.js#xfbml=1&version=v2.8";
+    fjs.parentNode.insertBefore(js, fjs);
+    }(document, 'script', 'facebook-jssdk'));
+</script>
+
```
psiemens: We don't need to include this script because we already load the Facebook JS SDK below.

psiemens (changes requested):
Hey @seanhsiung94, so it looks like this might be more complicated than I thought. 

The like/share buttons work on articles rendered by Django, but not in the AJAX-loaded articles that React inserts into the page. This is because the FB script that injects the buttons executes once when the page is initially loaded, but not when new articles are loaded below. 

I haven't tried this, but I think a possible solution would be to call `FB.XFBML.parse()` inside the `componentDidMount()` method of each `Article.jsx` component.

psiemens:
@seanhsiung94 can you update from `staging` and fix the conflicts? After that, it should pretty much be good to go!

PR #138, opened 2016-11-01
Removed "None" in image slideshow

### LINKED ISSUE #101: Empty captions showing 'None' in image slideshow
screen shot 2016-10-16 at 5 28 26 pm

If an image doesn't have a caption, the caption div is still rendered with the value 'None'. The caption div should only be rendered the image has a caption.

markoprodanovic:
Fixed #101 
Removed the “None” and added 20px of padding between picture and credit
if no caption

@psiemens could you take a look at this? Still new to JavaScript conventions. Is this an appropriate place to (conditionally) modify the style?

### DIFF
--- a/ubyssey/static/src/js/components/GallerySlide.jsx
+++ b/ubyssey/static/src/js/components/GallerySlide.jsx
@@ -1,5 +1,5 @@
 var GallerySlide = React.createClass({
-  render: function(){
+  render: function() {
     var slideStyle = { width: this.props.width };
     var imageStyle = { backgroundImage: "url('" + this.props.src + "')" };
 
--- a/ubyssey/static/src/js/components/GallerySlide.jsx
+++ b/ubyssey/static/src/js/components/GallerySlide.jsx
@@ -1,26 +1,36 @@
 var GallerySlide = React.createClass({
-    render: function(){
-        var slideStyle = { width: this.props.width };
-        var imageStyle = { backgroundImage: "url('" + this.props.src + "')" };
+  render: function(){
+    var slideStyle = { width: this.props.width };
+    var imageStyle = { backgroundImage: "url('" + this.props.src + "')" };
 
-        var caption = (<p className="slide-caption" dangerouslySetInnerHTML={{__html: this.props.caption}}></p>);
-	var credit = (<p className="slide-credit" dangerouslySetInnerHTML={{__html: this.props.credit}}></p>);
-        return (
-            <li className="slide" style={slideStyle}>
-                <div className="inner">
-                    <div className="image">
-                        <div>
-                            <div className="img" style={imageStyle}></div>
-                        </div>
-                    </div>
-                    <div className='slide-meta'>
-                        { this.props.caption ? caption : null }
-                        { this.props.credit ? credit : null }
-                    </div>
-                </div>
-            </li>
-        );
-    }
+    var caption = (
+      <p
+        className="slide-caption"
+        dangerouslySetInnerHTML={{__html: this.props.caption}}></p>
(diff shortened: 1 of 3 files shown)


ubyssey/static/src/js/components/GallerySlide.jsx
```diff
@@ -3,6 +3,10 @@ var GallerySlide = React.createClass({
         var slideStyle = { width: this.props.width };
         var imageStyle = { backgroundImage: "url('" + this.props.src + "')" };
 
+        if (!this.props.caption) {
+            $('.slide-credit').css('padding-top', '20px')
+        }
+
         var caption = (<p className="slide-caption" dangerouslySetInnerHTML={{__html: this.props.caption}}></p>);
 				var credit = (<p className="slide-credit" dangerouslySetInnerHTML={{__html: this.props.credit}}></p>);
```
psiemens: The indentation is a little off here -- can you fix it to be like the line above?

ubyssey/static/src/js/components/GallerySlide.jsx
```diff
@@ -14,7 +18,7 @@ var GallerySlide = React.createClass({
                         </div>
                     </div>
                     { this.props.caption ? caption : null}
-										{ this.props.credit ? credit : null}
+					{ this.props.credit ? credit : null}
```
psiemens: Same here -- fix indentation

ubyssey/static/src/js/components/GallerySlide.jsx
```diff
@@ -3,6 +3,10 @@ var GallerySlide = React.createClass({
         var slideStyle = { width: this.props.width };
         var imageStyle = { backgroundImage: "url('" + this.props.src + "')" };
 
+        if (!this.props.caption) {
+            $('.slide-credit').css('padding-top', '20px')
```
psiemens: Generally we try to avoid using jQuery to manipulate elements inside React components. When it comes to styles, React actually has built-in support for props- and state-driven style rules like so:

``` jsx
var style = this.props.caption ? {paddingTop: 20px} : null;
<p className="slide-credit" style={style} />
```

However, in this case I don't think we need to be mutating the element style using JavaScript. This is due to bad code that I wrote, but a more elegant solution would be to wrap both the caption and credit elements in a single container with `padding-top: 20px` defined in the stylesheet.

``` jsx
<div className='slide-meta'>
  { this.props.caption ? caption : null}
  { this.props.credit ? credit : null}
</div>
```

``` css
.slide-meta {
  padding-top: 20px;
}
```

ubyssey/static/src/js/components/GallerySlide.jsx
```diff
@@ -4,7 +4,7 @@ var GallerySlide = React.createClass({
         var imageStyle = { backgroundImage: "url('" + this.props.src + "')" };
 
         var caption = (<p className="slide-caption" dangerouslySetInnerHTML={{__html: this.props.caption}}></p>);
-				var credit = (<p className="slide-credit" dangerouslySetInnerHTML={{__html: this.props.credit}}></p>);
+	var credit = (<p className="slide-credit" dangerouslySetInnerHTML={{__html: this.props.credit}}></p>);
```
markoprodanovic: I keep having weird indentation issues after I commit.

psiemens: Ah, it's probably because some of this code was written with 4-space tabs instead of 2. No worries -- I'll fix it for consistency.

psiemens:
Looks good! Thanks a bunch @markoprodanovic 👍

PR #161, opened 2016-12-14
Update footer.html

jimmyhli:
Update current Year Automatically


psiemens:
Hey @HaoyuanLi --

Your solution definitely works, but I think it would be more elegant to render the date on the backend (using Python instead of JavaScript).

This can be done using Django's builtin `now` template function. Take a look at the header template for an example: https://github.com/ubyssey/ubyssey-dispatch-theme/blob/staging/ubyssey/templates/headers/main.html#L21

PR #164, opened 2017-01-05
Update footer.html

jimmyhli:
Automatic update to current year


ubyssey/templates/homepage/footer.html
```diff
@@ -21,7 +21,11 @@
               <li><a href="{% url 'page' 'advertise' %}">Advertise</a></li>
               <li><a href="{% url 'archive' %}">Archive</a></li>
             </ul>
-            <div class="copyright">&copy; 2015 The Ubyssey</div>
+            <div class="copyright">&copy;  
+		    <ul class="meta">
```
psiemens: We don't need the `ul` in here -- you should be able to just do:

`<div class="copyright">&copy; {% now "Y" %} The Ubyssey</div>`

Other than that, looks great!

PR #175, opened 2017-01-22
Added video play button

napon:
Preview of changes:
screen shot 2017-01-21 at 3 26 11 pm

### DIFF
--- a/ubyssey/static/src/styles/modules/article/_embeds.scss
+++ b/ubyssey/static/src/styles/modules/article/_embeds.scss
@@ -155,8 +155,30 @@
     display: none;
     top: 0;
   }
-  .video-launch:hover {
+  .play-button {
+    content: "";
+    display: block;
+    position: absolute;
+    top: 40%;
+    left: 50%;
+    margin: -15px auto auto -23px;
+    border-style: solid;
+    border-width: 30px 0 30px 46px;
+    border-color: transparent transparent transparent white;
+    z-index: 2;
+  }
+  .video-caption {
+    margin-top: 6px;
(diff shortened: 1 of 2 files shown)


blx (approved):
👍

psiemens:
This looks really good! However, there are a few things I want to change:

- I'm not sure I like the serif font for the caption. Can we try using the same font as the title?
- The layout breaks for long captions/titles (see screenshot for example). I think it would be better if the text filled the container instead of being cut off at 2/3.
- I'd prefer if the play button only changed colour when hovering directly over it.
- The play button also feels a little bare -- can we add a circle around it?
- The layout breaks on mobile devices. Maybe we can cut out the caption and/or make the text smaller?

screen shot 2017-01-26 at 11 54 48 am

**Mobile view:**

screen shot 2017-01-26 at 11 58 56 am


**I took a shot at changing the layout. Let me know what you think!**

screen shot 2017-01-26 at 11 51 43 am

napon:
@peijiading The mock up looks good and I agree with the points you listed.
> I'd prefer if the play button only changed colour when hovering directly over it

This might be more of a UI/UX thing but the area where the user is able to click is currently the entire wrapper around the video. If we move the hover effect to only the button, should the video still be playable when the user clicks on the wrapper? or should that be changed to only play the video when clicking directly on the play button?

> The play button also feels a little bare -- can we add a circle around it?

Sure! Might be good to also put some shadow on button and text so they are visible on a white background video.

Do you already have the mockup done in code? If so, you can commit what you have and I can make additional touchups to the design.

psiemens:
> Do you already have the mockup done in code?

Yep, I'll add the code ASAP.

psiemens:
Moved to #194

PR #177, opened 2017-01-23
Make query param parsing more robust in archive view (fixes #171)

### LINKED ISSUE #171: Bad query params cause 500 error in archive view
(casting to int fails)

Examples:
```
GET /archive/?section_id
GET /archive/?year=a

blx:


psiemens:
👍  many thanks!

PR #185, opened 2017-01-30
Removed Weekend Reading Side Column

tylernee:
Just deleted a few lines of code.

### DIFF
--- a/ubyssey/templates/homepage/base.html
+++ b/ubyssey/templates/homepage/base.html
@@ -44,13 +44,7 @@
                 </article>
                 {% endwith %}
             </div>
-            <div class="multi-zone cell right">
-                {% if day_of_week == 5 or day_of_week == 6 %}
-                    {{ components.multi_zone|safe }}
-                {% else %}
-                    {{ components.print_issue|safe }}
-                {% endif %}
-            </div>
+          
         </div>
     </div>
     <div class="homepage-section row">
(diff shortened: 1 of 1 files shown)


ubyssey/templates/homepage/base.html
```diff
@@ -44,13 +44,7 @@ <h2 class="headline"><a href="{{ article.get_absolute_url }}?ref=frontpage">{{ a
                 </article>
                 {% endwith %}
             </div>
-            <div class="multi-zone cell right">
-                {% if day_of_week == 5 or day_of_week == 6 %}
-                    {{ components.multi_zone|safe }}
-                {% else %}
-                    {{ components.print_issue|safe }}
```
psiemens, line 51: We actually want to keep this line:

`{{ components.print_issue|safe }}`

Sorry, should have been more clear in the issue. You can remove the `if/else` and just keep the `print_issue` line.

ubyssey/templates/homepage/base.html
```diff
@@ -92,15 +86,8 @@ <h4 class="headline">{{ article.headline|safe }}</h4>
                     {% endfor %}
                 </ul>
             </div>
-            <div class="multi-zone cell right">
-                {% if day_of_week == 5 or day_of_week == 6 %}
-                    {{ components.print_issue|safe }}
```
psiemens, line 97: Same here

ubyssey/templates/homepage/base.html
tylernee: Oh okay! So did you want me to add it back in or did you already?

psiemens: Can you add it back? Thanks!

PR #193, opened 2017-02-01
making an update

rahuljoban:


ubyssey/templates/homepage/base.html
```diff
@@ -9,6 +9,7 @@
     <div class="homepage-section frontpage-container">
         <div class="frontpage">
             <div class="cell left">
+              test
```
rahuljoban: this isn't quite right, can you fix it?

PR #195, opened 2017-02-02
Add comments for commands to run when facing specific errors

jumbosushi:
Just so other ppl won't have to google around as much as I did in the future.


psiemens:
Thanks @jumbosushi! This is really helpful

PR #219, opened 2017-02-16
61 advice opinion tabs

luceliu:
corrected links to advice and submit opinion pages


psiemens:
Thanks for the fix! @luceliu 

Just made a few small tweaks -- merging now

PR #228, opened 2017-02-20
60   add top 5 stories sidebar

b1s0ndev:


ubyssey/templates/article/default.html
```diff
@@ -48,6 +48,14 @@ <h1 class="o-headline o-headline--article">{{ article.headline|safe }}</h1>
...
         {% endif %}
           {% include 'objects/advertisement.html' with size='box' name='Box_A' id=3 article=article.id %}
+          <h2 class="block-title">Most popular this week</h2>
+          <ol class="article-list">
+          {% for a in top_articles %}
+              <li><h4 class="headline"><a href="{{ a.get_absolute_url }}">{{ a.headline|safe }}</a></h4></li>
+
+          {% endfor %}
+          </ol>
+
```
psiemens: Can you pull of this code inside a containing div? Something like this:

```html
<div class="o-sidebar-block">
    <h2 class="o-sidebar-block__title">Most popular this week</h2>
    <ol class="o-articles-list">
    {% for a in top_articles %}
        <li><h4 class="headline"><a href="{{ a.get_absolute_url }}">{{ a.headline|safe }}</a></h4></li>
     {% endfor %}
    </ol>
</div>
```

The `.o-sidebar-block` class names are just a new CSS convention we're trying to use.

psiemens (changes requested):
Looks good! 

However, I'd like to use new classes/styles for this component. Take a look at the comment I left on the code -- it adds new class `o-sidebar-block` that we can use to standardize future sidebar items. Right now you're using the styles from the homepage sidebar, but I'd like to do something different for articles.

You can add a file called `styles/objects/_sidebar_block.scss` to define the new styles. You'll have to import it into the main SASS file (e.g. like this: https://github.com/ubyssey/ubyssey-dispatch-theme/blob/staging/ubyssey/static/src/styles/main.scss#L34)

PR #232, opened 2017-03-01
173 new inline banner ad placement

ReggieLai:
I think this is it from the HTML side? I tried playing around with the
styling, but then I read that styling is better done using CSS.


ubyssey/templates/objects/advertisement-shop.html
```diff
@@ -0,0 +1,2 @@
+<img src="https://cloud.githubusercontent.com/assets/2547035/23276248/764ffdf0-f9be-11e6-89ae-621e42e2f9f3.png"
```
psiemens: I think it'd be good to wrap this inside a container `div`, something like this:

```html
<div class="o-advertisement o-advertisement--shop">
    <img src="https://cloud.githubusercontent.com/assets/2547035/23276248/764ffdf0-f9be-11e6-89ae-621e42e2f9f3.png" width="728" height="90" alt="sample-ad"/>
</div>
```

Then you can use CSS to add better styling!

PR #274, opened 2017-07-12
Event calendar

joncppl:


ubyssey/static/src/styles/main.scss
```diff
@@ -34,6 +34,11 @@
 @import 'components/section-page';
 @import 'components/homepage';
 
+@import 'modules/events/calendar';
+@import 'modules/events/event';
+@import 'modules/events/date_box';
+@import 'modules/events/event_box';
```
psiemens: I want to move towards using components instead of modules, so you can simply change `modules/events` to `components/events`

ubyssey/static/src/styles/modules/events/_calendar.scss
```diff
@@ -0,0 +1,178 @@
+$border                           : 1px dotted $color-border-gray-dark;
+$calendar-event-time-label-color  : #006fcc;
+
+$tiles-margin                     : 35px;
+
+#calendar {
```
psiemens: We try to avoid using IDs for css targeting. Can you do `.c-event-calendar` instead?

ubyssey/static/src/styles/modules/events/_calendar.scss
```diff
@@ -0,0 +1,178 @@
...
+  @media($bp-smaller-than-phablet) {
+    padding: 0 2px;
+
+    h1 {
+       padding: 0 5px;
+    }
+  }
+}
+
+.calendar-container {
```
psiemens: Each classname should follow the conventions from the [style guide](https://github.com/ubyssey/ubyssey-dispatch-theme/wiki/SASS-Style-Guide):

```scss
.c-event-calendar__container {
  // Structure
  width: 100%;
}
```

ubyssey/static/src/styles/modules/events/_calendar.scss
```diff
@@ -0,0 +1,178 @@
+$border                           : 1px dotted $color-border-gray-dark;
+$calendar-event-time-label-color  : #006fcc;
+
+$tiles-margin                     : 35px;
+
+#calendar {
+  // max-width: 1600px;
```
psiemens: Remove comment

ubyssey/static/webpack.dev.config2.js
```diff
@@ -0,0 +1,21 @@
+var webpack = require('webpack');
```
psiemens: Were you trying out a new webpack config?

ubyssey/templates/article/blank.html
```diff
@@ -7,7 +7,7 @@
 {% block content %}
     <main class="page">
         <div class="u-container u-container--content">
-          {{ page.get_html|safe }}
+          {{ page.html|safe }}
```
psiemens, line 10: I could be wrong, but you shouldn't have to add the `|safe` filter to this anymore

ubyssey/urls.py
```diff
@@ -29,6 +32,9 @@
     # Elections
     url(r'^elections/$', theme.elections, name='elections'),
 
+    url(r'^calendar/$', theme.events_calendar, name='calendar'),
```
psiemens: I think I'd prefer to use `/events` for the calendar URL. Then we can keep the route structure consistent and do `events/(?P<event_id>[0-9]+)/$`  below

psiemens (changes requested):

ubyssey/static/webpack.dev.config2.js
joncppl: oops. commited by accident. will revert.

ubyssey/templates/article/blank.html
joncppl: My understanding:
Safe means it doesn't need escaping. 

everything in Publishable.html is returned by content_to_html  and is carefully generated. 
I think escaping it would break the generated html?

Removing the safe tag makes no noticeable different however (ie. doesn't break the code embeds and such), so it can be removed.

Made me realize other escaping issues with the content editor in dispatch... which I'll bring up separately.

psiemens: You're right -- we definitely don't want to escape it. The difference is that we're now marking this as a "safe" string before it gets to the template:

https://github.com/ubyssey/dispatch/blob/develop/dispatch/apps/content/render.py#L16

ubyssey/views.py
```diff
@@ -442,6 +448,109 @@ def newsletter(self, request):
...
+            if year not in events_by_date:
+                events_by_date[year] = OrderedDict()
+            if month not in events_by_date[year]:
+                events_by_date[year][month] = OrderedDict()
+            if day not in events_by_date[year][month]:
+                events_by_date[year][month][day] = []
+
+            events_by_date[year][month][day].append(event)
+
+        return events_by_date
```
psiemens: To keep things organized, add the above helper functions under a new class `EventsHelpers` in this file: https://github.com/ubyssey/ubyssey-dispatch-theme/blob/staging/ubyssey/helpers.py

ubyssey/views.py
```diff
@@ -442,6 +448,109 @@ def newsletter(self, request):
...
+        events = events[:HARD_MAX]
+
+        return events
+
+    def organize_events_by_date(self, events):
+        events_by_date = OrderedDict()
+
+        for event in events:
+            start = event.start_time # Zulu
+            start = start.astimezone(timezone('America/Vancouver'))
```
psiemens: Do you have to convert the timezone in this step, or can it be done later? (i.e. in the template)

ubyssey/views.py
```diff
@@ -442,6 +448,109 @@ def newsletter(self, request):
...
+            'url': "%sevent/%s/" % (settings.BASE_URL, event.id)
+        }
+
+    def event_detail(self, request, event_id):
+        # TODO: find a way to select without putting event_id in url
+        # eg. slug generated from date & name to avoid collision
+        event = get_object_or_404(Event, pk=event_id)
+
+        if event.is_submission or not event.is_published:
+            raise Http404('Event could not be found.')
```
psiemens: I think it'd prefer to perform the lookup all in one step:

```python
try:
  event = EventsHelpers.get_event(pk)
except:
  raise Http404('Event could not be found.'
```

Then in the helper:

```python
return Event.objects.get(pk=pk, is_submission=False, is_published=True)
```

ubyssey/views.py
```diff
@@ -442,6 +448,109 @@ def newsletter(self, request):
...
+        return events
+
+    def organize_events_by_date(self, events):
+        events_by_date = OrderedDict()
+
+        for event in events:
+            start = event.start_time # Zulu
+            start = start.astimezone(timezone('America/Vancouver'))
+            year = start.year
+            month = start.month
```
psiemens: Call this `month_name` for clarity, or just pass it directly into the line below

ubyssey/views.py
```diff
@@ -442,6 +448,109 @@ def newsletter(self, request):
 
         return render(request, 'objects/newsletter.html', {})
 
+    def get_calendar_events(self, category=None, months=None, start=None, end=None):
+
+        events = Event.objects.all().order_by('start_time')
+        events = events.filter(is_submission=False)
+        events = events.filter(is_published=True)
```
psiemens: You can condense this:

```python

events = Event.objects
  .filter(is_submission=False, is_published=True)
  .order_by('start_time')

File: ubyssey/templates/events/event_box.html [hunk 13]
```diff
@@ -0,0 +1,31 @@
+<div class="c-event">
+  {% if event.image and not hide_image %}
+  <a href="/event/{{ event.pk }}/">
```
psiemens: You can us the url template tag to perform a reverse lookup for this route: 

```django
{% url 'event' event.pk %}`

File: ubyssey/templates/headers/main.html [hunk 14]
```diff
@@ -24,6 +24,7 @@
             <li><a href="{% url 'page' 'about' %}">About</a></li>
             <li><a href="{% url 'page' 'advertise' %}">Advertise</a></li>
             <li><a href="{% url 'page' 'volunteer' %}">Volunteer</a></li>
+            <li><a href="{% url 'page' 'calendar' %}">Calendar</a></li>
```
psiemens: Change "Calendar" to "Events"

psiemens (changes requested):

ubyssey/views.py
joncppl: this is so they get put in the right day in the sorted data structure.

ubyssey/helpers.py
```diff
@@ -164,3 +170,65 @@ def get_random_articles(n, section, exclude=None):
...
+
+        return events
+
+    @staticmethod
+    def organize_events_by_date(events):
+        events_by_date = OrderedDict()
+
+        for event in events:
+            start = event.start_time # Zulu
+            start = start.astimezone(timezone('America/Vancouver'))
```
psiemens: Would it be possible to use the timezone defined in settings.py? I'm just worried about hard-coding this.

ubyssey/helpers.py
```diff
@@ -164,3 +170,65 @@ def get_random_articles(n, section, exclude=None):
...
+        if category is not None and category != 'all':
+            events = events.filter(category__exact=category)
+
+        HARD_MAX = 100
+        events = events[:HARD_MAX]
+
+        return events
+
+    @staticmethod
+    def organize_events_by_date(events):
```
psiemens: Rename to `group_events_by_date`

ubyssey/templates/headers/main.html
```diff
@@ -24,7 +24,7 @@
             <li><a href="{% url 'page' 'about' %}">About</a></li>
             <li><a href="{% url 'page' 'advertise' %}">Advertise</a></li>
             <li><a href="{% url 'page' 'volunteer' %}">Volunteer</a></li>
-            <li><a href="{% url 'page' 'calendar' %}">Calendar</a></li>
+            <li><a href="{% url 'events' %}">Calendar</a></li>
```
psiemens: Change `Calendar` to `Events`

ubyssey/urls.py
```diff
@@ -40,8 +42,8 @@
     # Elections
     url(r'^elections/$', theme.elections, name='elections'),
 
-    url(r'^calendar/$', theme.events_calendar, name='calendar'),
-    url(r'^event/(?P<event_id>[0-9]+)/$', theme.event_detail, name='event'),
+    url(r'^events/$', events.events_calendar, name='events'),
```
psiemens: You can just name this method `events.calendar`

ubyssey/urls.py
```diff
@@ -40,8 +42,8 @@
...
+    url(r'^event/(?P<event_id>[0-9]+)/$', events.event_detail, name='event'),
```
psiemens: This can be `events.event`

psiemens (changes requested):

psiemens (approved):

psiemens:
@joncppl Feel free to merge after fixing the conflicts

PR #281, opened 2017-07-24
Event calendar layout

psiemens:
Here are my proposed changes to the calendar page layout. I borrowed styles from the archive page in an effort to make the calendar more consistent with the rest of the site.


joncppl (approved):
Looks good. The category sidebar is much better now.

PR #307, opened 2017-08-18
Event detail redesign

Axel-Jacobsen:


ubyssey/templates/components/event-detail.html
```diff
@@ -0,0 +1,36 @@
...
+  <div class="c-event__buttons">
+    {% if event.event_url %}
+      <a class="c-button" href="{{ event.event_url }}">
+        <i class="fa fa-facebook-square"></i>
+        <span>View on Facebook</span>
+      </a>
+    {% endif %}
+    {% if event.ticket_url %}
+      <a class="c-button " href="{{ event.ticket_url }}">
+        <i class="fa fa-ticket" style="vertical-align: middle;"></i>
```
joncppl: I think this style is redundant with `ubyssey/static/src/styles/components/_button.scss`

joncppl (commented):
Code looks good.

Just the one thing, the spacing on the icon of the ticket button still looks wider that the FB button.

Perhaps the two buttons could be a bit farther apart also.

joncppl (approved):

PR #316, opened 2017-08-24
Added contexts to event submit pages

Axel-Jacobsen:
Also fixed error catching in retrieving data from Facebook / UBC Events

### DIFF
--- a/ubyssey/events/views.py
+++ b/ubyssey/events/views.py
@@ -12,10 +12,26 @@ from ubyssey.events.forms import EventForm
 from ubyssey.events.models import Event
 
 def submit_landing(request):
-    return render(request, 'events/submit/landing.html')
+
+    context = {
+        'meta': {
+            'title': 'Submit an Event',
+            'description': 'Hosting an event on or off campus? Submit it to us and we\'ll feature it on our website!'
+        }
+    }
+
+    return render(request, 'events/submit/landing.html', context)
 
 def submit_success(request):
-    return render(request, 'events/submit/success.html')
+
+    context = {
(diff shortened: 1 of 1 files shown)


ubyssey/events/views.py
```diff
@@ -56,7 +72,16 @@ def submit_form(request):
...
+    context = {
+        'form': form,
+        'url_error': url_error,
+        'meta': {
+            'title': 'Submit an Event',
+            'description': 'Hosting an event on or off campus? Submit it to us and we\'ll feature it on our website!'
+        }
+    }
+
+    return render(request, 'events/submit/form.html', context)#{'form': form, 'url_error': url_error})
```
psiemens: Extra comment on this line

ubyssey/events/views.py
```diff
@@ -125,4 +150,4 @@ def get_host_from_url(url):
     if m:
         return m.group(1)
     else:
-        raise EventError('URL provided is not a valid Facebook event or UBC event url')
```
psiemens, line 128: What does this change do?

psiemens (changes requested):

ubyssey/events/views.py
Axel-Jacobsen: Ah if we raise the error, the `submit_form` method will stop executing when trying to pull Facebook data. if we return `None`, it will turn `url_error` to `True` and the submit form displays properly

psiemens: Gotcha, makes sense 👍

psiemens (approved):

PR #331, opened 2017-09-17
Solution for failure to install MySQL-python

bf777:
Provided solution for cases where ``MySQL-python`` doesn't install (with ``requirements.txt`` or by itself with ``pip``).


psiemens:
Thanks Brandon! 

My apologies, but these setup instructions are a bit out of date and probably won't work due to some recent changes to Dispatch. I just finished updating them [here](https://code.ubyssey.ca/getting-started/installation/mac.html) -- do you mind opening a PR into [this repo](https://github.com/ubyssey/volunteer/blob/master/getting-started/installation/mac.md) with your changes?

PR #346, opened 2017-09-25
Fix: Make entire event box container clickable (fixes #344)

### LINKED ISSUE #344: Entire event box should be clickable
Right now, event boxes only have links on the photo and title. Users should be able to click anywhere on the container to go to the event page.

screen shot 2017-09-25 at 12 32 35 pm

i-ron-y:
Close #344

### DIFF
--- a/ubyssey/static/src/styles/modules/events/_event.scss
+++ b/ubyssey/static/src/styles/modules/events/_event.scss
@@ -64,6 +64,7 @@
 
 // EventBox
 .c-event--box {
+  display: inline-block;
   // Border
   border: solid 1px $color-border-gray;
   box-shadow: 0 1px 1px 0 rgba(0, 0, 0, 0.05);
--- a/ubyssey/static/src/styles/modules/events/_event.scss
+++ b/ubyssey/static/src/styles/modules/events/_event.scss
@@ -64,7 +64,7 @@
 
 // EventBox
 .c-event--box {
-  display: inline-block;
+  display: block;
   // Border
   border: solid 1px $color-border-gray;
   box-shadow: 0 1px 1px 0 rgba(0, 0, 0, 0.05);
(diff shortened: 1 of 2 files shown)


ubyssey/templates/components/event-box.html
```diff
@@ -1,15 +1,13 @@
-<div class="c-event c-event--box{% if size %} c-event--{{ size}}{% endif %}">
+<div onclick=location.href="{% url 'event' event.pk.hex %}"; style="cursor: pointer;" class="c-event c-event--box{% if size %} c-event--{{ size}}{% endif %}">
```
psiemens: Instead of using `onClick`, it'd be better to change the `div` to an `a` tag and use `href` as you normally would.

psiemens (changes requested):

i-ron-y:
How's this looking, any suggestions?

ubyssey/templates/components/event-box.html
```diff
@@ -1,31 +1,31 @@
-<div class="c-event c-event--box{% if size %} c-event--{{ size}}{% endif %}">
-  {% if event.image and not hide_image %}
-  <a href="{% url 'event' event.pk.hex %}">
+<a href="{% url 'event' event.pk.hex %}">
```
psiemens: The only problem here is that `<a>` tags are `display: inline;` by default, so wrapping one around a block element might cause issues in certain browsers. I'd opt for something like this:

Convert the `div` to an `a` instead of wrapping it
```html
<a class="c-event c-event--box" href="{% url 'event' event.pk.hex %}">
  ...
</a>
```

```scss
.c-event--box {
  display: inline-block;
  ...
}

**Comment by i-ron-y** [comment]
Ahh, that makes sense! I have made the suggested changes.

File: ubyssey/static/src/styles/modules/events/_event.scss [hunk 3]
```diff
@@ -64,6 +64,7 @@
 
 // EventBox
 .c-event--box {
+  display: inline-block;
```
psiemens: This should actually be `display: block;`. The default for `a` tags is `inline-block`, but we want this to behave like a `div`, which has `block` positioning. 

```scss
.c-event--box {
  // Structure
  display: block;

psiemens (changes requested):

psiemens (approved):

PR #351, opened 2017-09-27
Fix: Event and ticket links always include protocol (fixes #345)

### LINKED ISSUE #345: Event Facebook and ticket links should always include a protocol
Some Facebook and ticket links are missing a protocol (`http` or `https`), which makes the browser treat them like links to other pages on ubyssey.ca

**Requirements:**
- Before saving an `Event`, check that the Facebook and ticket URLs contain a valid protocol. If not, add one before saving

i-ron-y:
Close #345

### DIFF
--- a/ubyssey/events/forms.py
+++ b/ubyssey/events/forms.py
@@ -47,6 +47,18 @@ class EventForm(ModelForm):
 
         facebook_image_url = self.data.get('facebook_image_url')
 
+        old_event_url = self.data.get('event_url')
+        old_ticket_url = self.data.get('ticket_url')
+
+        has_protocol_event_url = old_event_url.startswith('http://') or old_event_url.startswith('https://')
+        has_protocol_ticket_url = old_ticket_url.startswith('http://') or old_ticket_url.startswith('https://')
+
+        if not has_protocol_event_url:
+            event.event_url = "http://" + old_event_url
+
+        if not has_protocol_ticket_url:
+            event.ticket_url = "http://" + old_ticket_url
+
         if not event.image and facebook_image_url:
             event.save_image_from_url(facebook_image_url)
                 
--- a/ubyssey/events/forms.py
+++ b/ubyssey/events/forms.py
@@ -47,18 +47,6 @@ class EventForm(ModelForm):
 
         facebook_image_url = self.data.get('facebook_image_url')
 
-        old_event_url = self.data.get('event_url')
-        old_ticket_url = self.data.get('ticket_url')
-
-        has_protocol_event_url = old_event_url.startswith('http://') or old_event_url.startswith('https://')
-        has_protocol_ticket_url = old_ticket_url.startswith('http://') or old_ticket_url.startswith('https://')
-
-        if not has_protocol_event_url:
-            event.event_url = "http://" + old_event_url
-
-        if not has_protocol_ticket_url:
-            event.ticket_url = "http://" + old_ticket_url
-
         if not event.image and facebook_image_url:
             event.save_image_from_url(facebook_image_url)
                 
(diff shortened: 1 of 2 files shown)


ubyssey/events/forms.py
```diff
@@ -47,6 +47,18 @@ def save(self, commit=True):
...
+
+        has_protocol_event_url = old_event_url.startswith('http://') or old_event_url.startswith('https://')
+        has_protocol_ticket_url = old_ticket_url.startswith('http://') or old_ticket_url.startswith('https://')
+
+        if not has_protocol_event_url:
+            event.event_url = "http://" + old_event_url
+
+        if not has_protocol_ticket_url:
+            event.ticket_url = "http://" + old_ticket_url
+
```
psiemens: This logic is good, but I think a helper function would make things a bit cleaner:

```py
def format_url(url):
  # code goes here

event.event_url = format_url(event.event_url)

**Review by psiemens — CHANGES_REQUESTED** [review]
This is great, but an even more robust approach would be to place this code in a `pre_save` hook for the `Event` model.

Here's an example: https://github.com/ubyssey/ubyssey.ca/blob/develop/ubyssey/events/models.py#L77

File: ubyssey/events/models.py [hunk 2]
```diff
@@ -91,6 +91,35 @@ def send_submitted_email(sender, instance, **kwargs):
...
+    return not (url and url.strip())
+
+def has_protocol(url):
+    return url.startswith('http://') or url.startswith('https://')
+
+def has_protocol_or_empty(url):
+    return has_protocol(url) or url_empty(url)
+
+def format_url(url):
+    if has_protocol_or_empty(url):
```
psiemens: I don't think you need to use a helper in this case. I'd be okay with something like this:

```py
if has_protocol(url) or not url:
  return
```

ubyssey/events/models.py
```diff
@@ -91,6 +91,30 @@ def send_submitted_email(sender, instance, **kwargs):
...
         instance.save()
 
+
+@receiver(pre_save, sender=Event)
+def format_event_ticket_url(sender, instance, **kwargs):
+    urls_formatted = ( ( has_protocol(instance.event_url) or not instance.event_url ) and
+                       ( has_protocol(instance.ticket_url) or not instance.ticket_url ) )
+    
+    """Add "http://" in front of (non-empty) event_url and ticket_url if protocol missing."""
+    if instance.is_submission and not urls_formatted:
```
psiemens: In this case, you don't have to check if the instance is a submission. I think it's safe to say that the URLs should be sanitized whenever the event is being saved.

Furthermore, I don't think the `urls_formatted` check is necessary. You should be able to just do this without the `if` statement:

```py
instance.event_url = format_url(instance.event_url)
instance.ticket_url = format_url(instance.ticket_url)
```
If the url is already formatted, the above code won't modify it.

You should also be able to omit `instance.save()` because the instance is about to be saved anyways.

ubyssey/events/models.py
```diff
@@ -91,6 +91,30 @@ def send_submitted_email(sender, instance, **kwargs):
...
+
+        instance.event_url = format_url(instance.event_url);
+        instance.ticket_url = format_url(instance.ticket_url);
+
+        instance.save()
+
+def has_protocol(url):
+    return url.startswith('http://') or url.startswith('https://')
```
psiemens: This looks good! Nice and simple.

psiemens (changes requested):

ubyssey/events/models.py
i-ron-y: You're absolutely right... I'm not sure what I was doing wrong the other day because when I tested it without some sort of check, I had gotten an error about infinite recursion or something to that effect. But I tried it again right now and everything's fine!

psiemens: Awesome!

It's likely that the call to `instance.save()` was creating a loop by triggering the `pre_save` hook again.

psiemens (approved):

PR #353, opened 2017-09-28
Ubc event cron

Axel-Jacobsen:
Here is the Cron job, albeit a little late. I believe that it should work, but it is a little difficult to test.

### DIFF
--- a/ubyssey/events/management/commands/import_events.py
+++ b/ubyssey/events/management/commands/import_events.py
@@ -108,3 +108,14 @@ class Command(BaseCommand):
 
         feedObj.create_new_events()
         feedObj.remove_old_events()
+
+        return feedObj.feed.status
+
+def main():
+
+    feedObj = UBCEventsRSSFeed('http://services.calendar.events.ubc.ca/cgi-bin/rssCache.pl?days=2&mode=rss')
+
+    feedObj.create_new_events()
+    feedObj.remove_old_events()
+
+    return feedObj.feed.status
--- a/ubyssey/events/management/commands/import_events.py
+++ b/ubyssey/events/management/commands/import_events.py
@@ -20,6 +20,7 @@ class UBCEventsRSSFeed(object):
         """The new events must be added to both the Event Model and ScrapedEvent"""
 
         new_events = self.get_new_events()
+        num_events = 0
 
         for event in new_events:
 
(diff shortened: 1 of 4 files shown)


ubyssey/events/management/commands/import_events.py
```diff
@@ -108,3 +109,16 @@ def handle(self, **options):
...
+        return feedObj.feed.status
+
+def main():
+
+    feedObj = UBCEventsRSSFeed('http://services.calendar.events.ubc.ca/cgi-bin/rssCache.pl?days=2&mode=rss')
+
+    feedObj.create_new_events()
+    feedObj.remove_old_events()
+
+    http_status = feedObj.feed.status
```
psiemens: I'd prefer to have a cleaner interface for the `UBCEventsRSSFeed`. Something along these lines:

```py
feed = UBCEventsRSSFeed('http://services.calendar.events.ubc.ca/cgi-bin/rssCache.pl?days=2&mode=rss')

updated = feed.update()

if updated:
  status = SOME_VALUE
else:
  status = SOME_OTHER_VALUE

File: ubyssey/events/management/commands/import_events.py [hunk 2]
```diff
@@ -108,3 +109,16 @@ def handle(self, **options):
...
+
+    return HttpResponse(status=http_status)
```
psiemens: Would it be possible for this response to also include a short summary of the results? Maybe a JSON payload:

```json
{
  "added": 12,
  "removed": 3
}

File: ubyssey/events/urls.py [hunk 3]
```diff
@@ -9,5 +9,6 @@
     url(r'^submit/$', views.submit_landing, name='events-submit-landing'),
     url(r'^edit/success/', views.edit_success, name='events-edit-success'),
     url(r'^edit/(?P<secret_id>.+)', views.edit, name='events-edit'),
+    url(r'^import/', views.event_import, name='events-import'),
```
psiemens, line 12: `views.event_import` can just be `views.import`

psiemens (changes requested):
Looks good! Added some suggestions.

Do you think it would make sense to remove the manage command now that it has its own URL? I just find it a little confusing having the `main` method inside `import_events.py`

ubyssey/events/urls.py
Axel-Jacobsen: wouldn't having a method called `import` conflict with the `import` keyword in python?

ubyssey/events/management/commands/import_events.py
Axel-Jacobsen: I dont think i quite understand the purpose of this; I made some changes and pushed so take a look?  I think I completed the goal of this code snippet

psiemens (approved):

PR #398, opened 2017-11-30
388 three rows of articles

victorhyde:
Section pages now have 3 rows of articles instead of 1

### DIFF
--- a/ubyssey/views/main.py
+++ b/ubyssey/views/main.py
@@ -210,7 +210,9 @@ class UbysseyTheme(object):
             'type': 'section',
             'featured_articles': {
                 'first': featured_articles[0],
-                'rest': featured_articles[1:4]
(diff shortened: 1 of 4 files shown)


ubyssey/views/main.py
```diff
@@ -210,7 +210,9 @@ def section(self, request, slug=None):
             'type': 'section',
             'featured_articles': {
                 'first': featured_articles[0],
-                'rest': featured_articles[1:4]
+                'row1': featured_articles[1:4],
+                'row2': featured_articles[4:7],
+                'row3': featured_articles[7:10]
```
psiemens, line 215: I just discovered this, but Django has a `slice` template tag that would allow you to do this slicing directly in the template: https://docs.djangoproject.com/en/1.11/ref/templates/builtins/#slice

PR #400, opened 2017-12-07
Display latest print issue on home page

### LINKED ISSUE #393: Allow support for print issue box on homepage

i-ron-y:
Close #393


ubyssey/views/main.py
```diff
@@ -54,6 +56,14 @@ def home(self, request):
...
         title = '%s - UBC\'s official student newspaper' % self.SITE_TITLE
 
+        # Scraping the URL and cover image URL of Ubyssey's latest print issue
+        page = requests.get('http://search.issuu.com/ubyssey/docs/recent.rss')
+        soup = BeautifulSoup(page.content, 'xml')
+        latestIssue = soup.findAll('item')[0]
+
+        issuu_url = latestIssue.find('link').get_text()
+        issuu_img = latestIssue.find('media:content').get('url')
+
```
psiemens: This is shockingly simple!

However, I don't think it's feasible to make a request to Issuu on every page load. At the very least it will slow our site down, and at worst they might block our IP address. 

I have an idea for a solution that would require a bit of a change to Dispatch, but would result in only having to make one request to the Issuu RSS feed and then caching the results. 

Notice how the events sidebar widget defined [here](https://github.com/ubyssey/ubyssey.ca/blob/develop/ubyssey/widgets.py#L35-L72) has a `context` method. This is simply an override of the `context` base method defined in the `Widget` base class: https://github.com/ubyssey/ubyssey.ca/blob/develop/ubyssey/widgets.py#L35-L72

By default, `context` just returns the `data` parameter unchanged. But if the user decides to override the method, they have the freedom to update the data and add/change parameters before it is passed to the template.

I'm thinking it would be great to add an additional *optional* base class called `before_save` that has the same functionality, but allows users to mutate the data before it is saved to the database. It would only get called when the widget is saved. This is where you could place the scraping code above to prevent it from being called too many times.

Thoughts?

i-ron-y: True, I was worried about bombarding Issuu with requests and was thinking we probably have to store the url and cover image on our own server somewhere...

That's a good idea. Should we make the print issue box a widget (I don't think it current is)? And in which table would we be saving the Issuu url and cover image?

i-ron-y:
Argh, pardon the mess... the last three commits are the ones relevant to this PR.

Thanks again for implementing the before_save hook, Peter!

PR #406, opened 2018-01-10
created Twitter widget

rithikha:
Created (twitter widget) #340

### DIFF
--- a/ubyssey/static/src/styles/modules/widgets/_facebook_video_big.scss
+++ b/ubyssey/static/src/styles/modules/widgets/_facebook_video_big.scss
@@ -17,12 +17,10 @@
 
 .c-facebook-video__title-container {
 	display: flex;
-  font-weight: 800;
+	font-weight: 800;
 	margin-bottom: 1rem;
 	padding-left: 1rem;
 	padding-right: 1rem;
-
-
 
 	.c-facebook-video__title {
 		flex: 1;
@@ -30,7 +28,7 @@
 		font-weight: $font-weight-bold;
 		padding-left: 1rem;
 		padding-right: 0rem;
-		font-family: 'ub-lft-etica', 'LFT Etica', 'Helvetica Neue', 'Helvetica', 'Arial', sans-serif;
+		font-family: $font-default;
 		margin-top: 1rem;
 		margin-bottom: 0rem;
 	}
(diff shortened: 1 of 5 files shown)


ubyssey/static/src/styles/modules/widgets/_facebook_video_big.scss
```diff
@@ -15,6 +15,29 @@
 	}
 }
 
+.c-facebook-video__title-container {
+
+	display: flex;
+  font-weight: 800;
```
jumbosushi: Missing spacing here

ubyssey/widgets.py
```diff
@@ -115,6 +115,43 @@ def context(self, result):
         return result
 
 @register.widget
+<<<<<<< Updated upstream
```
jumbosushi: Still have git merge leftovers in this file haha

ubyssey/static/src/styles/modules/widgets/_twitter.scss
```diff
@@ -0,0 +1,4 @@
+.c-twitter-widget-container {
+height: 30rem;
```
jumbosushi: Need spacing here

ubyssey/static/src/styles/modules/widgets/_facebook_video_big.scss
```diff
@@ -15,6 +15,29 @@
...
+	margin-bottom: 1rem;
+	padding-left: 1rem;
+	padding-right: 1rem;
+
+
+
+	.c-facebook-video__title {
```
jumbosushi: Spacing is bit different from others in here

jumbosushi (commented):
Great start! There's a django bug when I run the site with this which I'll look into, but let's try to fix these styles first

ubyssey/widgets.py
psiemens: This could be what's causing the server to crash

ubyssey/templates/widgets/frontpage/twitter-front-page.html
```diff
@@ -0,0 +1,5 @@
+<div class="c-twitter-widget-container">
+<a class="twitter-timeline" href="https://twitter.com/ubyssey"></a>
```
psiemens, line 2: This shouldn't be hardcoded -- you can return the `{{ handle }}` value that is saved by the widget.

psiemens (changes requested):

PR #408, opened 2018-01-16
Add CSS for printing articles

jumbosushi:
- Fix article print
- Hide ads
- Hide social modules
- Hide infinite scroll articles

### DIFF
--- a/ubyssey/static/src/styles/modules/_advertisements.scss
+++ b/ubyssey/static/src/styles/modules/_advertisements.scss
@@ -110,9 +110,8 @@ main.article .o-advertisement--leaderboard {
 }
 
 // Hide ads when printed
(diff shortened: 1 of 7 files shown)


ubyssey/static/src/styles/modules/_advertisements.scss
```diff
@@ -108,3 +108,11 @@ main.article .o-advertisement--leaderboard {
     margin-right: auto;
   }
 }
+
+// Hide ads when printed
+div[class^="o-advertisement"] {
```
psiemens: I think this should just be:

```scss
.o-advertisement {
```

jumbosushi: yea that's more safe

PR #426, opened 2018-03-01
Add election widget

jumbosushi:
WIP: fb live div needs more love on mobile
edit: good to go


jumbosushi:
FB live widget should be decoupled from election widget later on

PR #430, opened 2018-03-08
Get rid of the &:after part of .o-article-embed--code in _embeds.scss

### LINKED ISSUE #429: Weird spacing in about partnership section
Need to investigate where this is coming from

emptyspot

i-ron-y:
Close #429 

The problem in #429 seems to have come about because the width for code embeds is set to 100%, and then the &:after part adds a top padding to ensure a 16:9 proportion for the embed. This basically results in a lot of extra width added to the embed, and then a lot of top padding to keep up with the extra width.


jumbosushi:
Amazing worked locally. Great work!

PR #448, opened 2018-05-22
Preview page fix

JamieRL:
Change get_page helper to how get_article helper works. This fixes issue where you cannot preview a page that is not published.


psiemens (approved):
👍

PR #450, opened 2018-05-25
Streamline release process

JamieRL:
Pass version in with context. I added a global context processor but it seems to not work everywhere unless I pass it with the main.py views' context.


psiemens (approved):
Looks good to me. Just need to resolve the conflicts

JamieRL:
I'm going to run a double check on this as I'm not sure that the version is making it to every page.

PR #452, opened 2018-05-28
Change get_years in article helper

JamieRL:
'get_years' now gets all the unique years from the 'published_at' field for articles where 'is_published' is true


Rowansdabomb (approved):
looks good

PR #461, opened 2018-06-06
Get version from template tags

JamieRL:
Now store the version in settings.py. Templates access it through template tag. dispatch_tags in dispatch handles building the static files urls


Rowansdabomb (approved):
Great

PR #478, opened 2018-06-14
remove old dispatch dependency

JamieRL:
remove dependency on old dispatch api wrapper. Moved functionality into new dispatch api.


Rowansdabomb (approved):
Looks good to me

Issue #479, opened 2018-06-14
update advertise page


JamieRL:
In this section add a new box that says 'Article' (with the associated ad dimensions)
Clicking this box should pan the viewer to the right and show a mockup article page with the available ad slots

![image](https://user-images.githubusercontent.com/5742577/41435920-1fa2c65c-6fd5-11e8-8991-98292a7af61f.png)

JamieRL:
Completed, waiting for ad update to be finished.

JamieRL:
Completed in #481

PR #481, opened 2018-06-15
479 advertise new

JamieRL:
create new advertise page to for our new ad slots/sizes


JamieRL:
Addressing issue #479

Rowansdabomb (approved):
Everything seems to be in order

PR #511, opened 2018-07-10
490 section columns

JamieRL:
Views and templates for columns, the rest is on dispatch.


JamieRL:
I'm wondering if there should be a way to hide columns. Currently when you make a column it will always show up on that section page. But there's no way to make the column not appear unless you delete it.

ubyssey/views/main.py
```diff
@@ -274,6 +277,62 @@ def section(self, request, slug=None):
         t = loader.select_template(['%s/%s' % (section.slug, 'section.html'), 'section.html'])
         return HttpResponse(t.render(context))
 
+    def subsection(self, request, slug=None):
+        try:
+            subsection = Subsection.objects.get(slug=slug)
```
Rowansdabomb: add is_active

Rowansdabomb (approved):
Looks good to me

PR #521, opened 2018-07-30
520 event save error

JamieRL:
- Swap ordering in format_url() to not call has_protocol when url is None.


Rowansdabomb (approved):
looks good

PR #528, opened 2018-08-02
492 push notifications

JamieRL:


ubyssey/static/src/js/notifications.js
```diff
@@ -0,0 +1,89 @@
...
+  const outputArray = new Uint8Array(rawData.length);
+
+  for (let i = 0; i < rawData.length; ++i) {
+    outputArray[i] = rawData.charCodeAt(i);
+  }
+  return outputArray;
+}
+
+const cookieName = 'notification_subscription'
+
```
Rowansdabomb, line 25: we should hold all the cookie names in a file somewhere else

ubyssey/static/src/js/notifications.js
```diff
@@ -0,0 +1,89 @@
...
+}
+
+function setCookie(uuid) {
+  Cookies.set(
+    cookieName,
+    {uuid: uuid},
+    { path: '/' }
+  )
+}
+
```
Rowansdabomb, line 45: we use these cookie functions in several places, we should make them generalized and move them into a cookies/helpers.js

PR #539, opened 2018-08-29
add only the newest 3 subsection articles in section page

JamieRL:
- add limit to get published articles so that it only displays 3 newest articles


Rowansdabomb (approved):
looks good

PR #541, opened 2018-08-31
replace background-image with img tag

Rowansdabomb:


ubyssey/static/src/styles/modules/events/_event.scss
```diff
@@ -9,8 +9,8 @@
 .c-event__image {
   // Structure
   position: relative;
-  padding-top: 50%;
-
+  // padding-top: 50%;
```
Rowansdabomb: delete background props and comments

JamieRL (approved):
Approved.

PR #542, opened 2018-09-10
Subsection redesign

JamieRL:
Redesign/rework for subsections. Including some styling changes and changes for how the articles to be shown for each subsection are selected.

1. Added a small header that lists all of the active subsections that have published articles
header_subsections

2. Only show 1 subsection on the section page. This subsection is chosen by taking the subsection (associated with this section) that has the most recently published article. 

3. So as to still promote new articles when they are published, the featured articles section is reverted to its original functionality, however the articles shown in the featured subsection are now chosen as to not show any duplicate articles that are already on the page.
duplicates

### DIFF
--- a/ubyssey/helpers.py
+++ b/ubyssey/helpers.py
@@ -6,7 +6,7 @@ from django.http import Http404
 from django.db import connection
 from django.db.models.aggregates import Count
 
-from dispatch.models import Article, Page, Section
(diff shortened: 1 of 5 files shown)


Rowansdabomb (approved):
Looks good to me, Complicated query for subsections but what are you gonna do

PR #565, opened 2018-10-01
Replace `get_medium_url ` with `get_square_url`

kellylawrence:
This should retrieve the square thumbnail sized post image.


Rowansdabomb (changes requested):
What is `get_square_url`? I can not find any instance of this function.

kellylawrence:
@Rowansdabomb My bad. I was looking at the following starting at [line 466](https://github.com/ubyssey/dispatch/blob/develop/dispatch/modules/content/models.py#L466). I'll submit an update shortly using `get_thumbnail_url`.

```
SIZES = {
    'large': (1600, 900),
    'medium': (800, 600),
    'square': (250, 250)
}
```

Rowansdabomb (approved):
Looks good

PR #568, opened 2018-10-02
Update README.md

jimmyhli:
Fixed broken links


jumbosushi:
Ah good catch. Thanks for the fix!

Issue #555, opened 2018-09-20
Fix the cookie disclaimer


Rowansdabomb:
We have a cookie disclaimer to comply with EU regulations. If you are interested, you can read more about it [here](https://www.cookiebot.com/en/cookie-consent-banner/?gclid=CjwKCAjwio3dBRAqEiwAHWsNVStXbF6jN047UlDgYtOPerVmcliqi-P2Tz9fy7wJDrd9Ag3lFK_N0hoCb8QQAvD_BwE).

Our header is positioned fixed over the cookie disclaimer, partially obscuring it. 

![image](https://user-images.githubusercontent.com/19273530/45841035-77dba100-bccd-11e8-971b-f28829a13ac1.png)

Let's ensure that our cookie disclaimer is always fully visible.

**Steps to get started**
You can find the sass stylesheet for the cookie disclaimer [here](https://github.com/ubyssey/ubyssey.ca/blob/develop/ubyssey/static/src/styles/components/_cookie-disclaimer.scss)

The Cookie Disclaimer is actually implemented as a React component. You probably won't need to change the component itself, but for the sake of completeness you can find the source code for the CookieDisclaimer.jsx file [here](https://github.com/ubyssey/ubyssey.ca/blob/develop/ubyssey/static/src/js/components/CookieDisclaimer.jsx)

You can either move the cookie disclaimer so it sits below the header. Or you may position the disclaimer to render at the bottom of the screen. Ensure your fix accounts for **both the desktop and mobile** styles of the site. 

**Note**
Because the cookie disclaimer's visibility is dependent on a cookie itself, You may want to disable cookies while working on this. 

If you are testing on google chrome, You can also delete the cookie by entering the browser inspector `ctrl-shift-i` or right click in the browser and select `inspect` from the drop down. Then, in the inspector window, navigate to the Application -> Cookies -> https://www.ubyssey.ca. Find the Cookie with the name `cookie_disclaimer` and delete it's **value**. You will likely have to do this multiple times.

kellylawrence:
@Rowansdabomb Would there be any issues with setting `<div id="cookie-disclaimer"></div>` in the [base.html template](https://github.com/ubyssey/ubyssey.ca/blob/909a1cb1064c4320d25e1dcf850f29c2a5cf5a09/ubyssey/templates/base.html#L105) as the last rendered element? (After `<div id="content-wrapper">`)

[...]

PR #569, opened 2018-10-02
#555 Fix cookie disclaimer placement

kellylawrence:
This should resolve issue #555.


jumbosushi:
Thanks for the PR! Looks good to me. CC: @Rowansdabomb

PR #571, opened 2018-10-06
563 podcasts page

Rowansdabomb:
This PR is not ideal, there is no built in pagination for episodes and such. However, Sam is eager to release podcasts so I'm hoping to get this initial release out soon.


psiemens:
Looks good!

I have some feedback regarding the code style, but if you keep this branch open after merging I can address those changes later (in order to get this published).

_settings/settings-local.py
```diff
@@ -24,6 +24,7 @@
 
 DEBUG = True
 USE_TZ = True
+LOCAL_DEV = True
```
jumbosushi: Like the idea of setting flag for different environment. Not a blocker, but I wonder if this will be more explicit if do something like `ENV = 'development' and `ENV = 'production'`

jumbosushi (approved):

PR #572, opened 2018-10-12
Fix breaking news

JamieRL:
change filter for breaking news to use timezone.now instead of deprecated datetime.datetime.now


Rowansdabomb (approved):
LGTM

PR #573, opened 2018-10-12
fix trending news style issue for homepage

Rowansdabomb:
This was an issue with using the font tag, which is no longer supported in html5.

### DIFF
--- a/ubyssey/templates/widgets/frontpage/default.html
+++ b/ubyssey/templates/widgets/frontpage/default.html
@@ -8,7 +8,7 @@
         {% if articles.trending %}
           <div class="o-trending-article">
           {% with articles.trending as trending_article %}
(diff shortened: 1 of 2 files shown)


jumbosushi (approved):
:100:

ubyssey/templates/widgets/frontpage/default.html
```diff
@@ -8,10 +8,10 @@
         {% if articles.trending %}
           <div class="o-trending-article">
           {% with articles.trending as trending_article %}
-              <div>
+          <span>
```
jumbosushi: Not a blocker but indent is bit off here

Rowansdabomb: I will fix it before I merge

PR #576, opened 2018-10-12
create a dev options scheme in settings

Rowansdabomb:
This is an idea for local settings dev options, more verbose than a single LOCAL_DEV tag. What do you think? Maybe a  dict instead of an array?

### DIFF
--- a/_settings/settings-local.py
+++ b/_settings/settings-local.py
@@ -22,34 +22,50 @@ MIDDLEWARE = (
     'django_user_agents.middleware.UserAgentMiddleware',
 )
 
+TIME_ZONE = 'America/Vancouver'
+
 DEBUG = True
 USE_TZ = True
 
-TIME_ZONE = 'America/Vancouver'
+DEV_OPTIONS = ['LOCAL_DOCKER', 'LOCAL', 'PRODUCTION']
+DEV_ENV = DEV_OPTIONS[0]
 
-################ LOCAL MYSQL ##################
-# DATABASES = {
-#     'default': {
-#         'ENGINE': 'django.db.backends.mysql',
-#         'NAME': 'ubyssey',
-#         'USER': 'root',
-#         'PASSWORD': 'ubyssey',
-#         'HOST': '127.0.0.1',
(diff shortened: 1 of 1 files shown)


_settings/settings-local.py
```diff
@@ -22,34 +22,50 @@
...
-#     },
-# }
+DEBUG = True
+USE_TZ = True
 
+DEV_OPTIONS = ['LOCAL_DOCKER', 'LOCAL', 'PRODUCTION']
+DEV_ENV = DEV_OPTIONS[0]
+
+###############  PRD MYSQL  #################
+if DEV_ENV == DEV_OPTIONS[2]:
```
jumbosushi, line 34: Don't think defining prd config in local file since this file is [not being loaded](https://github.com/ubyssey/ubyssey.ca/blob/develop/.travis.yml#L9) by travis

_settings/settings-local.py
```diff
@@ -22,34 +22,50 @@
...
-#         'PASSWORD': 'ubyssey',
-#         'HOST': '127.0.0.1',
-#         'PORT': '3306',
-#     },
-# }
+DEBUG = True
+USE_TZ = True
 
+DEV_OPTIONS = ['LOCAL_DOCKER', 'LOCAL', 'PRODUCTION']
+DEV_ENV = DEV_OPTIONS[0]
```
jumbosushi, line 31: Maybe we can provide in a comment available options what the available settings are. `ENVIRONMENT` might be more clear about what this actually means
```python
# ['LOCAL_DOCKER', 'LOCAL', 'PRODUCTION']
ENVIRONMENT = 'LOCAL_DOCKER'
```

_settings/settings-local.py
Rowansdabomb: This is for PRD database connection, ie when you want to proxy into the database

_settings/settings-local.py
Rowansdabomb: Yea I like `ENVIRONMENT` better

PR #581, opened 2018-10-25
Allow pdb to be attach on docker

jumbosushi:
Based on [stackoverflow answer like this ](https://stackoverflow.com/questions/32403664/make-pdb-work-with-docker), we need to add two lines to allow docker to let terminal attach to ongoing `pdb` session. I've had it on my local file, but it should be updated on repo level to allow everyone to debug easily (added in our [faq](http://code.ubyssey.ca/faq) page)


Rowansdabomb (approved):
Sweet, PDB FTW!

PR #587, opened 2018-10-30
fix cookie disclaimer

Rowansdabomb:


jumbosushi:
LGTM :100: : 
Non blocker, but ould be nice if there's also a screenshot / desc of what this is fixing :+1:

Issue #588, opened 2018-10-31
Cookies do not persist


Rowansdabomb:
Cookies do not seem to persist past the browser session. 

It looks like the cookie expiry dates are past dated to 1969, this is probably the source of the issue.

![image](https://user-images.githubusercontent.com/19273530/47824910-51793e80-dd2c-11e8-9a48-aae7ae400d04.png)

jimmyhli:
This `YSC` cookie is set by youtube.com though, so I doubt if there's anything we can do on our end. I believe this `YSC` simply stands for `Youtube Session Cookie` (though I can't find anything credible to back my assumption), and is designed to be deleted at the end of a browser session. 😃 

I've seen some sources that indicate that this YSC has something to do with sharing videos and stuff, but I don't know about that for sure.

Rowansdabomb:
Sorry if that image is unclear, `YSC` is not the cookie in question. The only cookies we are concerned about currently are the `cookie_disclaimer` and `ubyssey_site_visit` cookies.

jimmyhli:
Oh, sorry. `YSC` caught my attention because it was highlighted 🤣 

Do you mind taking a look at https://github.com/ubyssey/ubyssey.ca/pull/591

Rowansdabomb:
fixed in #591

PR #591, opened 2018-11-03
update the validity of cookies

jimmyhli:
This PR updates the validity of the two types of cookies as mentioned in https://github.com/ubyssey/ubyssey.ca/issues/588

The original code also specifies the `path` as `'/'`, which i believe is probably the default setting anyways, so it might not be necessary, I didn't delete it though.


Rowansdabomb (approved):
Looks good

PR #592, opened 2018-11-04
Simplify cookie disclaimer logic

jumbosushi:
## Issue

- Noticed that `cookieNames.js` and `helper.js` in `/components/Cookies` were not being used
- `cookie_disclaimer` name is bit ambiguous
- Value of `cookie_disclaimer` is stored as a json string when we only use it as a string

image

## Solution

- Remove unused files + now import files via `/components/Cookies/index.js` file that imports both `AdblockSplash` and `CookieDisclaimer` components
- Rename `cookie_disclaimer` to `ubyssey_cookie_disclaimer`
- Simplify cookie disclaimer logic to use string instead of manipulating JSON in string format 

image


Rowansdabomb (approved):
Lgtm

PR #609, opened 2018-11-20
Update manifest + update service-worker

jumbosushi:
Enables add to home screen + adds basic caching for our main js files


service-workers/service-worker.js
```diff
@@ -1,3 +1,56 @@
...
+  '/static/js/article-' + VERSION + '.js',
+  '/static/images/ubyssey-logo.svg'
+];
+
+self.addEventListener('install', function(event) {
+  // Perform install steps
+  event.waitUntil(
+    caches.open(CACHE_NAME)
+      .then(function(cache) {
+        console.log('Opened cache');
```
Rowansdabomb: Do you want this to log in production?

Rowansdabomb (approved):
Non-blocker console.log. LGTM

service-workers/service-worker.js
```diff
@@ -1,3 +1,56 @@
+var VERSION = '1.4.80'
+var CACHE_NAME = 'ubyssey-cache-v' + VERSION;
+var urlsToCache = [
+  '/static/js/vendors-' + VERSION + '.js',
+  '/static/js/main-' + VERSION + '.js',
+  '/static/js/article-' + VERSION + '.js',
+  '/static/images/ubyssey-logo.svg'
+];
```
Rowansdabomb, line 92: I think we should additionally cache:
- all the .woff font files. 
- main-version.css
- blockadblock-version.js
- dfp-version.js

Rowansdabomb (changes requested):
Just a few more things I think we can cache with this PR. So stoked to get this running, hopefully it makes a noticeable difference for page load times!

Rowansdabomb (approved):
LGTM

PR #617, opened 2018-12-02
limit max instances, reduce idle instances, scale up cpu utilization …

Rowansdabomb:
…before new instances


psiemens (approved):
Looks good! Curious to see how this affects the deployment

PR #620, opened 2018-12-11
Change local middleware to < 1.10 syntax

jumbosushi:
Mini PR. Django < 1.10 use the `MIDDLEWARE_CLASSES` syntax and  > 1.10 syntax is `MIDDLEWARE`. We're importing [dispatch's setting file](https://github.com/ubyssey/dispatch/blob/develop/dispatch/default_settings.py) that still uses `MIDDLEWARE_CLASSES` syntax, and all other setting file seems to be in this syntax for now so we should match it on local setting file.


Rowansdabomb (approved):
LGTM

Rowansdabomb (changes requested):
Could we update the version to 1.4.127 so we don't lose the `settings-local.py` from the version updates in the future

Rowansdabomb (approved):
LGTM

PR #632, opened 2019-01-10
Add django-debug-toolbar to requirements.txt

jumbosushi:
We forgot to add it to the requirements.txt so new volunteers need to manually install it themselves rn


Rowansdabomb (commented):
LGTM

PR #637, opened 2019-02-25
upgrade to react 14

Rowansdabomb:
Upgrades the ubyssey.ca to react v14

Updated package.json including webpack to v3. Ran jscodeshift to automagically update some breaking changes. 

Since we do not have any frontend js tests, I tested this by removing all errors and warnings in the console. I have tested the gallery component, polls component, and the suggested articles component. 

You will need to rerun `yarn install` from your ubyssey/static directory to test this.


jumbosushi (approved):
:100:

Issue #646, opened 2019-03-28
Upgrade Gulp to v4.0.0


jumbosushi:
## Issue

GitHub has been warning us about the security risk of `minimatch` npm package 
![image](https://user-images.githubusercontent.com/9669739/55197487-fda24280-516f-11e9-9f4a-fb59268c2ac1.png)

### Set up

New `package.json` is pushed at `upgrade-gulp-v4` branch

**All of this npm work is done under `ubyssey.ca/ubyssey/static` dir**

Go to the dir, and run `yarn`, them make sure gulp version is v4 by running `gulp -v`

![image](https://user-images.githubusercontent.com/9669739/55198074-cd5ba380-5171-11e9-85d7-89576662cfcd.png)
 

Test `gulpfile.js` by running `gulp`
### Identifying issue

Looking at current`yarn.lock` file, some of the `minimatch` [version is in v2](https://github.com/ubyssey/ubyssey.ca/blob/develop/ubyssey/static/yarn.lock#L1687). 

Package that use outdated `minimatch` are following:
* `glob-stream`
* `glob`
* `globule`

By using [yarn's why command](https://yarnpkg.com/en/docs/cli/why) (ex. `yarn why glob-module`) the dependency comes from current Gulp version that we use which is v3.9.1

### Current Output

Think we need to re-write gulp task definitions? 

![image](https://user-images.githubusercontent.com/9669739/55197925-58886980-5171-11e9-84ef-7fbb34c178eb.png)

### Suggested Steps

Check out [Gulp's Documentation](https://gulpjs.org/API.html#gulp-task-name-fn) as well as [this migration guide](https://www.liquidlight.co.uk/blog/how-do-i-update-to-gulp-4/)

PR #650, opened 2019-04-09
Upgrade to gulp v4

razvannesiu:
This PR solves (#646 ) the update to gulp v4.

Rewrote the gulpfile.js according to the new standards of gulp v4.


Rowansdabomb:
I'm noticing that this is not passing the gulp build test. This might not be an issue with the upgrade to gulp as much as an issue with how Travis handles the test, but it needs to be addressed.

Rowansdabomb (changes requested):
Could you please remove the package-lock.json

Rowansdabomb (approved):
LGTM

Issue #656, opened 2019-05-07
Upgrade webpack from 3.12.0 to 4.30.0


razvannesiu:
It is probably time to do an update for webpack. If you go inside _/ubyssey.ca/ubyssey/static_ you can run `npm outdated` and it shows that there a few packages that are outdated (webpack is one of them).
Here is one article that talks about the migration: https://medium.com/practo-engineering/migrating-to-webpack-4-for-beginners-e52e49753235

PR #657, opened 2019-05-07
Upgrade to webpack v4

razvannesiu:

It solves the issue (#656 ) with the upgrade to webpack v4.

Created a webpack.common.js file that contains the shared config, and then the webpack.prod.js and webpack.dev.js that extend the shared config file. Updated the gulpfile.js to import the new webpack files.


ghost:
lgtm

Rowansdabomb (approved):
LGTM

PR #665, opened 2019-05-09
updated python code in ubyssey.ca code base with cloned repositiory

ghost:

update python from 2.7 to 3

ran 2-to-3 automator and cloned repository than forked and removed yarn.lock


Rowansdabomb (approved):
This looks ok to me

Issue #601, opened 2018-11-15
Separate Video Section


jumbosushi:
We have awesome centennial video (shout out to Marina), and we want a place to showcase it.
Currently, it needs to be integrated into another article but would be great if it could be we can have our own Video page 

![image](https://user-images.githubusercontent.com/9669739/48580398-f28efa00-e8d3-11e8-832d-8f7bf6ab240b.png)

PR #674, opened 2019-05-29
601 add videos for ubyssey

razvannesiu:
This solves issue #601 
- adds thumbnail images of the video to the front page article previews
- adds a video section to the site header
- adds a video page 

Built a paginated video page in main.py. Conditionally check featured_media is an image or video when rendering preview images. Added some helper dispatch_filters to get youtube video ids from video urls.


ghost:
works on my end

PR #675, opened 2019-06-08
Django update to v2.2.2 LTS

razvannesiu:

This PR solves the issue of updating Django to v2.2 which is an LTS version. We should probably do another major update in 2022.

I followed the release notes for Django 2.0, 2.1, 2.2 and I made the changes.
https://docs.djangoproject.com/en/2.2/releases/2.2/


ghost:
LGTM

PR #676, opened 2019-06-27
Add video section on frontpage

razvannesiu:

This PR solves the issue of not having the video section on the frontpage.

Added a VideoHelper in helpers.py to handle urls, section_videos.html in base.html, column.html, featured.html files. Created a video object in the main.py file in the context object of the home function.


ghost:
looks good to me

ghost (approved):
looks good to me

Issue #663, opened 2019-05-09
Searching archive for people returns 2 more results


razvannesiu:
## What behavior were you expecting?

![Screenshot (91)](https://user-images.githubusercontent.com/11470906/57489777-f2acf880-726b-11e9-9a1d-63874f88b0cc.png)

I was expecting to see only 4 results for "Rowan". 

## What actually happened?

There are 6 results in total for "Rowan" (which 2 more than the actual people count). 

## Steps to reproduce

1. Go to archive.
2. Search "Rowan" or any other person.
3. Observe the bug.

## What was your environment like?

Chrome 74.0.3729.131 on Windows 10.

PR #682, opened 2019-08-30
Add images, videos, podcasts, podcast episodes on the archive

razvannesiu:

It solves the issue of not having images, videos, podcasts, podcast episodes on the archive. Also, it displays images, articles, podcasts, podcast episodes for each editor, in the archive. This PR fixes the issue (#663 ) of having a wrong count of objects displayed on the archive. 

I added a list.html template for each media object that is displayed on the archive. To solve the issue with the wrong count of objects displayed in the archive, I computed the start index and page of each set of media objects (ie: the start index and page of images, the start index and page of articles, etc.).

### DIFF
--- a/ubyssey/helpers.py
+++ b/ubyssey/helpers.py
@@ -362,7 +362,7 @@ class VideoHelper(object):
         return "%svideos/" % (settings.BASE_URL)
 
     @staticmethod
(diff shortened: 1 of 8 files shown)


ghost (approved):
archive works fine. LGTM.

PR #698, opened 2020-03-12
Modify dockerfile: install npm on container, move -y flag on nodejs i…

keeganland:
697 - docker-compose up error message

Added an ``apt-get install`` to the Dockerfile to install npm on docker container

### DIFF
--- a/local-dev/Dockerfile
+++ b/local-dev/Dockerfile
@@ -5,9 +5,10 @@ ENV PYTHONUNBUFFERED 1
 RUN export DEBIAN_FRONTEND=noninteractive
 RUN apt-get update \
   && apt-get -y install build-essential curl \
-  && apt-get -y install vim\ 
+  && apt-get -y install vim \ 
   && curl -sL https://deb.nodesource.com/setup_6.x | bash - \
-  && apt-get install -y nodejs \
+  && apt-get -y install nodejs \
+  && apt-get -y install npm \
   && nodejs -v \
   && npm -v
 COPY ./ubyssey.ca ./ubyssey.ca


local-dev/Dockerfile
```diff
@@ -5,9 +5,10 @@ ENV PYTHONUNBUFFERED 1
 RUN export DEBIAN_FRONTEND=noninteractive
 RUN apt-get update \
   && apt-get -y install build-essential curl \
-  && apt-get -y install vim\ 
+  && apt-get -y install vim \
```
ghost, line 8: why is vim removed?

local-dev/Dockerfile
```diff
@@ -5,9 +5,10 @@ ENV PYTHONUNBUFFERED 1
 RUN export DEBIAN_FRONTEND=noninteractive
 RUN apt-get update \
   && apt-get -y install build-essential curl \
-  && apt-get -y install vim\ 
+  && apt-get -y install vim \ 
   && curl -sL https://deb.nodesource.com/setup_6.x | bash - \
-  && apt-get install -y nodejs \
+  && apt-get -y install nodejs \
```
ghost, line 10: why did you switch order of install and -y?

local-dev/Dockerfile
keeganland: I didn't. Just put a space after vim in the version I happened to commit

local-dev/Dockerfile
keeganland: Due to someone's suggestion that npm's absence and our need to install it "manually" was due to nodejs installing improperly at this line, because typically an installation of nodejs brings npm with it. It was suggested I try testing to see if "apt-get -y install nodejs" instead made a difference. It did not seem to do so, but I left it like this for consistency of formatting

PR #704, opened 2020-04-22
new version 1.6.28 with mainmaller bottom fix issue font

davidkao0902:
Mainmaller goes to the bottom and the same font size for issues.
New version 1.6.28.


ghost (approved):
LGTM

PR #707, opened 2020-05-27
Ad widget and header changes

davidkao0902:
Added Advertisement widget for front page zone,
Changed header spacing for homepage,
Small changes on advertisement spacing.
Updated quote box with new styling


keeganland (approved):
lgtm, though put a pin in it to think of a maybe tidier solution for the same things

PR #709, opened 2020-05-29
Version ssot ized

keeganland:
Version variables haven't been getting the version from a single source of truth

Created `version.txt` as a single source of truth, updated every variable that had been manually updated with version releases to read from this file


davidkao0902 (approved):
Good stuff

PR #713, opened 2020-06-03
Settings and Dispatch revision

keeganland:

Setting files styles were very out of date and depended very strongly on Dispatch. 
We've dramatically updated how Dispatch works, and made corresponding changes here.

New settings app which uises Dispatch's new DispatchConfig class


davidkao0902 (approved):
Seems good, could add back DEBUG = FALSE back to production setting

PR #714, opened 2020-06-04
Upgrade django

keeganland:

Django out of date for revised production environment

Upgraded Django


davidkao0902 (approved):
Great

PR #717, opened 2020-06-09
fix frontpage headline format and updated issuu widget

davidkao0902:
-headline now set to 4 lines max
     -if exceed the maximum 4 lines. There will be ... in the end of the 4th line.
-Updated the featured issuu to May 26th and included the picture as a hyperlink as well.


keeganland (approved):
lgtm

PR #720, opened 2020-06-17
Config portability

keeganland:

Lack of portability of project due to hardcoded configs

Refactored config settings into .env, which lives in Google Cloud Secrets Manager


davidkao0902 (approved):
lgtm

PR #725, opened 2020-06-18
magazine 2019-2020 front page style update, set preset issuu values t…

davidkao0902:
- Updated magazine front page
- Set preset issue values to current issues.


keeganland (approved):
lgtm

PR #726, opened 2020-06-19
update styling for magazine to support mobile

davidkao0902:
- Updated support for mobile for magazine
- Made sections more visible


keeganland (approved):
lgtm

PR #737, opened 2020-07-10
Inline sql refactor

keeganland:

Very messy inline SQL in view

Refactors to Django's ORM


davidkao0902 (approved):
pretty cool lgtm

ubyssey/helpers.py
```diff
@@ -7,6 +8,8 @@
 from django.conf import settings
 from django.http import Http404
 from django.db import connection
+import django.db.models
```
psiemens, line 11: This line imports all definitions from `django.db.models`, but the line below imports specific members. I would remove this line in favour of the one below.

ubyssey/helpers.py
```diff
@@ -76,66 +79,44 @@ def insert_ads(content, article_type='desktop'):
...
-            TIMESTAMPDIFF(DAY, published_at, NOW()) <= %(max_days)s as age_deadline
-            FROM dispatch_article
-        """
-
-        query_where = """
-            WHERE head = 1 AND
-            is_published = %(is_published)s AND
-            parent_id NOT IN (%(excluded)s)
-        """
+    def get_frontpage(section=None, sections=[], exclude=[], limit=7, is_published=True, max_days=14):
```
psiemens, line 82: Maybe the `section` parameter should just be removed, seeing as it's semantically the same as `sections=[section]`?

psiemens (commented):
This is awesome, great work!

I wonder if there's a way to make it _even more clear_ what this query is doing? For example, do you think this reads any clearer, or does it just move the logic further away?

```py
Case(
  reading_time_is('morning', start='morning_start', then=1.0, else=0.0),
  reading_time_is('midday', start='midday_start', end='midday_end', then=1.0, else=0.0),
  ...
)

PR #739, opened 2020-07-14
remove section argument from get_frontpage

keeganland:
unnecessary arguments in get_frontpage

removed the one instance of `section` being used, removed it entirely from get_frontpage method


davidkao0902 (approved):
lgtm

PR #740, opened 2020-07-15
Inline sql subsection helper

keeganland:

Subsection helper method used inline SQL

Rewrote inline SQL to Django ORM


davidkao0902 (approved):
lgtm

PR #745, opened 2020-07-18
Staticfiles revision to manifest static files storage

keeganland:

Version numbers needed manual updating when it is possible for Django to automatically version static files

Use Django's ManifestStaticFilesStorage feature


davidkao0902 (approved):
lgtm

PR #747, opened 2020-07-23
Manifest static contd

keeganland:

Odd layout of collected static files

seperated /dist/ into its own folder according to django namespacing conventions, /static/ubyssey/


davidkao0902 (approved):
Everything worked as intended! Good stuff

PR #748, opened 2020-07-23
Magazine 500 fix

keeganland:

static() call that refers to nonexistant staticfile

point it at an existing staticfile


davidkao0902 (approved):
lgtm

PR #759, opened 2020-08-05
set upcoming-widget to be created at 1200px width

davidkao0902:
- set container to be created at above 1200px


keeganland (approved):
lgtm

PR #760, opened 2020-08-05
fix mobile fw-story banner issue and update some suggested article an…

davidkao0902:
- fix mobile fw-story banner issue
- update suggested article styling
- update ubyssey.ca footer styling


keeganland (approved):
lgtm

PR #764, opened 2020-08-07
deleted comments and prints; added restriction when resizing the full…

davidkao0902:
- added condition for desktop fw-story template
- remove comments in main.js


keeganland (approved):
lgtm

PR #765, opened 2020-08-10
Mixin refactor

keeganland:
proof of concept for major refactor that gets rid of helper classes, uses mixins, and organizes views according to standard Django expectations


ubyssey/mixins.py
```diff
@@ -0,0 +1,276 @@
+import datetime
+from datetime import datetime
```
psiemens, line 3: Are both of these imports needed?

ubyssey/mixins.py
```diff
@@ -0,0 +1,276 @@
...
+
+from django.utils import timezone
```
psiemens: I'd group this with the other Django imports below

ubyssey/views/main.py
```diff
@@ -16,39 +16,70 @@
...
 from ubyssey.helpers import ArticleHelper, PageHelper, SubsectionHelper, PodcastHelper, NationalsHelper, FoodInsecurityHelper, VideoHelper
+from ubyssey.mixins import ArticleMixin
 
 def parse_int_or_none(maybe_int):
     try:
         return int(maybe_int)
     except (TypeError, ValueError):
         return None
 
+class UbysseyHomePageView(ArticleMixin, TemplateView):
```
psiemens: 👏 

Nice, I think it makes a lot of sense to use a class-based view for the homepage.

ubyssey/mixins.py
```diff
@@ -0,0 +1,276 @@
...
+from django.db import connection
+from django.db.models import Case, ExpressionWrapper, DurationField, F, FloatField, OuterRef, Subquery, Value, When 
+from django.db.models.aggregates import Count
+
+from dispatch.models import Article, Page, Section, Subsection, Podcast, Image, ImageAttachment
+
+from ubyssey.events.models import Event
+
+
+class ArticleMixin(object):
```
psiemens, line 21: Great job consolidating all of the article-related logic. A few points:

- Did you consider using dependency injection rather than a mixin? Mixins are a powerful tool for inheritance, but in this case it's just being used to provide additional functionality to a class. Having this functionality be implicitly mixed in with the the view logic (e.g. in `UbysseyHomePageView`) feels like it could get a bit confusing.
- I think it might be worthwhile to split this class up even further in the future. Some methods use the database and some don't, but overall it's quite a bit of logic that might benefit from being moved into several smaller classes.

psiemens (commented):
Nice! This is a huge step in the right direction 🎉

ubyssey/mixins.py
keeganland: I defaulted to mixins, though I did hesitate, because it's so common to see code that has to be reused with View classes in particular take the form of mixins. It at least seemed a step up from having this code in a helper class. I agree it still needs work.

Wasn't sure about dependency injections because I've found Django's kind of opaque about the lifetime of the View object. I don't know when the service object would have to be instantiated.

Definitely agreed about splitting it up further. It will probably retain some stuff but other stuff it was lumped with originally (not my doing) seems like it really ought to be abstracted into its own thing.

ubyssey/mixins.py
keeganland: I think this is an artefact of datetime having a submodule _also called_ datetime and my thinking I needed both and that I didn't want to type "datetime.datetime" a bunch because that looks really odd. 

It's occurred to me though that maybe my use of both didn't survive into the committed code. I'll give it a check

PR #830, opened 2021-04-19
Design fixes

mwhite98:
After some feedback, I've done a few small design changes. The main one is fixing the header so that it no longer bleeds into the body on mobile. Also added back in google analytics stuff.


keeganland (approved):
lgtm

PR #848, opened 2021-05-11
Streamblock templates added

LinsamuelATM:

Added stream block templates for each stream block widget.


keeganland (approved):
lgtm

PR #863, opened 2021-05-27
updated digital print issues on the homepage

LinsamuelATM:
Updated the most recent print issues on the homepage


keeganland (approved):
lgtm

Issue #967, opened 2022-02-18
Videos page - not on homepage & authors not showing


ihciem:
- Videos "section" not on home page

- Authors not showing up on Videos page <img width="1440" alt="Screen Shot 2022-02-17 at 16 08 32" src="https://user-images.githubusercontent.com/45521929/154592711-d2982227-edad-47a2-8ae8-464fbd5d4443.png">

ihciem:
Videos page authors: DONE

PR #969, opened 2022-02-24
Videospage authors

ihciem:

Video authors now showing on videos page and linked to each author's page.
[#967](https://github.com/ubyssey/ubyssey.ca/issues/967)

Updated video model with function get_authors_string to avoid logic in template. 
Function is called in template to get all authors and link them.
Function borrowed/modified from ArticlePage model.


keeganland (approved):
lgtm

Issue #940, opened 2022-02-04
Images stretching on homepage


ihciem:
## What actually happened?
<img width="851" alt="Screen Shot 2022-02-04 at 9 28 50 AM" src="https://user-images.githubusercontent.com/45521929/152596970-fed64918-16f0-4aab-8a73-f25df4750a28.png">

keeganland:
This is likely caused by someone setting the wagtail resize rules incorrectly in some templates

See: https://docs.wagtail.org/en/stable/topics/images.html

PR #974, opened 2022-03-02
height now adjusts automatically

ihciem:

Fixes issue #940 

Wagtail resize rules for width only was being overridden by css, but didn’t adjust height, which was causing the weird ratios.

Image height now set to auto.
This fix was documented to cause layout problems but we do not seem to have any at the moment.
[https://www.smashingmagazine.com/2020/03/setting-height-width-images-important-again/](https://www.smashingmagazine.com/2020/03/setting-height-width-images-important-again/)


keeganland (approved):
LGTM

PR #975, opened 2022-03-02
Fixed height issues for featured articles imgs

ihciem:

#940 

Height set to auto for other featured articles on homepage.


keeganland (approved):
lgtm

PR #976, opened 2022-03-04
Font stack update

ihciem:
Updated font stack to 

- Headline: Noto Serif Display
- Byline: Noto Sans
- Body text: Cormorant Garamond (medium)

Includes support for woff2, woff, ttf, and svg (in that order).

Screen Shot 2022-03-03 at 18 25 00

Screen Shot 2022-03-03 at 18 25 41


keeganland (approved):
LGTM

PR #978, opened 2022-03-08
Homepage image resizing problem

ihciem:

Decreased quality of images in featured articles on homepage (frontpage).

Screen Shot 2022-03-07 at 19 46 43

Centre image increased to width of 715 (approx/greater what should actually load).
Image fill increased to 250 by 250 for column on left (approx/greater than what should actually load).
So rendition phase of image captures better quality for use.


keeganland (approved):
lgtm

Issue #982, opened 2022-03-09
Full-width article byline not appearing


brittkhat:
Is this a bug, or feature disabled for the stock layout template full-width? Requires investigation.

PR #984, opened 2022-03-10
Changed function in template to correct one

ihciem:

#982 

Changed function used in template.


keeganland (approved):
LGTM

PR #987, opened 2022-03-31
Github protocol update

ihciem:

Docker image building runs into problems due to requirements.txt

git:// changed to https://
See [https://github.blog/2021-09-01-improving-git-protocol-security-github/](https://github.blog/2021-09-01-improving-git-protocol-security-github/)
See also: [StackOverflow discussion/fixes](https://stackoverflow.com/questions/70663523/the-unauthenticated-git-protocol-on-port-9418-is-no-longer-supported)

### DIFF
--- a/home/templates/home/home_page.html
+++ b/home/templates/home/home_page.html
@@ -28,12 +28,13 @@
 
     <div class="homepage-section row">
       <div class="sections" >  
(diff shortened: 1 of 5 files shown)


keeganland (approved):
lgtm

Issue #1001, opened 2022-05-17
Ad Behavior - home page and base


brittkhat:
Review home_page.html and base.html for lines where ad space and slugs are implemented. Debug and determine how code is functioning, use inspect tools firefox and scope out missing ad slots on homepage. If necessary rewrite ad inventory lines 40-43 and starting on line 56 article_list_sidebar_block.

Purpose: to ensure proper cpm are forecast in ad manager, all ad slots are displaying as expected, and code base is properly implemented.

ihciem:
Ad manager settings tweaked. Ad slots weren't activated properly, etc. Closing.

PR #1005, opened 2022-05-26
1001 - above cut ad slot fix

keeganland:

#1001 (in part)

There were two uses of the same id due to templating error. Earlier Dispatch templates made it unclear that two elements required distinct IDs.

User could select ad slot before for sidebar, this adds ad selection to above cut block


ihciem (approved):
lgtm

PR #1008, opened 2022-05-30
Added command chaining in Dockerfile

Danisa-R:
Optimizes container performance by reducing run commands

Added command chaining in Dockerfile to optimize container performance


keeganland:
This is something that we had once upon a time, but the problem with it is when Docker would error, it would associate the error with a single RUN command. And there was knowledge loss within the Ubyssey such that no one knew how to fix any Docker errors. And it used to error _a lot_.

Commands were unchained and associated with distinct RUN commands because it would get considerably easier to tell exactly _where_ the problems occurred, when they inevitably would.

The point of Docker in our environment is to get a consistent development environment up and running. Being able to respond to new volunteers' inevitable problems is a far more important requirement than having a meticulously optimized Dockerfile— the Dockerfile isn't actually being used in the context of say, Kubernates or anything.

Perhaps this optimization doesn't necessarily cut against the requirement we be able to respond efficiently to volunteers' errors, but _prima facie_ it appears to do so. And it seems to sacrifice this benefit for another benefit that we frankly don't care about— consistency matters much more than speed (is speed a huge problem? I haven't heard any complaints about it). I need more convincing that this a change we actually want rather than a change for the sake of change.

PR #1016, opened 2022-07-14
remove buggy problematic filter for article ads

keeganland:

Once upon a time, article pages would run a filter to give ad div ids highly customized ids which would then be picked out of the HTML document by javascript when registering ads. We've dramatically simplified ads registration so now this filter causes bugs.

Just straight up removed the filter and all reference to it.


ihciem (approved):
lgtm

PR #1052, opened 2022-11-09
Article page changes

ihciem:
1. HTML hierarchy changes: moved Article page media to under "article-content" 
2. Made edits to CSS for right-image full-width Articles as requested


brittkhat (approved):
lgtm

PR #1068, opened 2023-03-03
cardblock html and py additions for magazine 2023

brittkhat:
Created textless_banner_block.html, mag_2023_card.html, and modified blocks.py to create cardblocks for the magazine


keeganland (approved):
LGTM

PR #1080, opened 2023-04-27
homepage and article page image quality fix

ihciem:

Image quality issues on the homepage and the main photos within indiv article pages.

Images were being loaded onto the page at one size [through Wagtail](https://docs.wagtail.org/en/stable/topics/images.html) and then resized through CSS (e.g. width: auto) to another, bigger size, causing the image quality issues.

Fixed through adjusting loaded sizes through Wagtail template.


brittkhat (approved):
lgtm

PR #1082, opened 2023-05-08
Add the image of the top left article to the homepage

### LINKED ISSUE #972: Home page request: show top left article image on frontpage
- **images** for articles on homepage should be displayed (top article on left)

FireBoyAJ24:

When the homepage is rendered the above cut stream html renders the first few articles. Most to all times the <article class="secondary padded"> is rendered first for the first article. Within this tag there is no <div class="image"> tag, therefore, no image is rendered to the top left article of the homepage. To fix the issue I have added an image div which is the same as the article tag with class thumb row padded to the article tag with class secondary padded.

Things we can also look at,
1. Are there any article tags that is loaded for the first top left article than secondary padded.
2. Do we want to change the structure of the article tag secondary padded?
3. Do we want to change the size of the image to make the image fit with the homepage layout?

### DIFF
--- a/home/templates/home/stream_blocks/above_cut_block.html
+++ b/home/templates/home/stream_blocks/above_cut_block.html
@@ -14,6 +14,16 @@
 				{% with articles.1 as article %}
 				{% if article %}
 				<article class="secondary padded">
+					<div class="image">
+						<a href={% pageurl article %}>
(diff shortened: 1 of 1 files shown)


SamuelmdLow:
This is made irrelevant by the redesign

PR #1083, opened 2023-05-09
1041 feature suggested articles akshanjay

### LINKED ISSUE #1041: Feature: suggested articles
Break down into several smaller updates.

Implement to side of Article pages

FireBoyAJ24:

#1041 

1. I have finished adding suggested articles sidebar.
2. The user will see a maximum of 6 recent articles from the section the reader is reading in.
3. I have added the functions needed to get the suggested articles and finished the remaining HTML to make the article page more presentable.
4. The format of the suggested articles now comes from @ihciem. After reviewing her work I have added the condition, that if the article page is in the guide then no featured suggested articles are presented.
5. The suggested articles will first come from the category that the host article is from. If there is no articles then the recent articles from the section is presented. Instead of @ihciem 4 articles, I have increased it to 6 articles.


SamuelmdLow (changes requested):
The suggested articles list should be be its own template and placed in the article with `{% include` That way it will be easier to edit independently and move around. Also since it is exactly the same as the sidebar in the home page I think you should make a new `sidebar` app, move the sidebars templates and blocks to that, and then reuse that template for the suggested articles.

Also since this pulled from visual-essay, I fetched and merged the updated visual essay branch

FireBoyAJ24:
> The suggested articles list should be be its own template and placed in the article with `{% include` That way it will be easier to edit independently and move around. Also since it is exactly the same as the sidebar in the home page I think you should make a new `sidebar` app, move the sidebars templates and blocks to that, and then reuse that template for the suggested articles.
> 
> 
> 
> Also since this pulled from visual-essay, I fetched and merged the updated visual essay branch


Having a separate sidebar app would allow for abstraction. However, with the different ways at which the sidebar is presented in each type of article page, it makes it difficult to make the abstraction.

I will further investigate the issue and make amendments accordingly.

SamuelmdLow:
I moved the logic out of the template and simplified a few things

SamuelmdLow (approved):

PR #1084, opened 2023-05-09
980 ams election samlow

### LINKED ISSUE #980: AMS Template
Create a template for AMS articles for 2023. Referendum menu to be above div containing feature image. Template available under select stock layout in wagtail. Idea: have the menu be derived from any h2 with a id created in the article.

SamuelmdLow:

fixed issue #980 

Created a page links block and a header block. Clicking a page link will send the user to the header which shares the id


FireBoyAJ24 (changes requested):
Can we add all the inline styles and the scripts to be in external style sheets and scripts?

article/templates/article/stream_blocks/pageLink.html
```diff
@@ -0,0 +1,19 @@
+{% load static %}
+
+<div class="header-menu">
+     <ul>
+          {% for item in self.list %}
+               <style>
+                    .pageLink{{item.id}} {
+                         background-color: #{{item.colour}} !important;
```
FireBoyAJ24: Is there anyway to avoid !important?

FireBoyAJ24 (approved):
The !important is a small aspect and will hopefully not impact the rest of the code base but it is best avoided unless all options have been dried up.

article/templates/article/stream_blocks/pageLink.html
SamuelmdLow: I've removed it. Turns out it wasn't actually important

SamuelmdLow:
Merged into redesign

PR #1085, opened 2023-05-11
Moved the above cut block into the sections div

### LINKED ISSUE #1040: Homepage design: vertical sectioning
Make the sections of the homepage extend the third, right hand column.

i.e. column: homepage_above_cut_box (Box A) (immediately continues to the next ad or the Blog section, or the Print Issues section)

Before:
Screen Shot 2022-10-13 at 16 21 22

After (crudely edited, but here's the rough idea):
Screen Shot 2022-10-13 at 16 21 22 edited

SamuelmdLow:

Does what is requested in issue #1040

The above cut block is moved into the sections div so that the sidebar continues up the right side column all the way. I also changed the heading on above the cut to "Most recent" instead of "home" because it better describes what the block displays.

### DIFF
--- a/home/templates/home/home_page.html
+++ b/home/templates/home/home_page.html
@@ -37,14 +37,15 @@
 	Additional fields/blocks will go here to alert students to some special events
 	{% endcomment %}
 
-	<!-- Above the cut content. Likely the first things the user sees when going to the site, before scrolling down -->
-	{% for block in self.above_cut_stream %}
-	{% include_block block %}
-	{% endfor %}
-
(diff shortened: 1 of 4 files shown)


FireBoyAJ24 (changes requested):
Extra space is needed between the "Most Recent" and the horizontal nav bar. In addition, can we change the title from "Most Recent" to "Recent Posts" or "Recent Articles"?

The way you can increase the top margin is by going to main.scss (I believe) and changing the h1.section:first-child class margin-top.

SamuelmdLow:
The new front page design that I'm working makes this pull request redundant

PR #1086, opened 2023-05-11
Removed old UA google tag and replaced it with the GA4 tag

SamuelmdLow:

Universal analytics is ending service in the beginning of July so we have to migrate to Google analytics 4

I replaced the old UA google tag with the new GA4 tag


brittkhat:
lgtm

brittkhat (approved):
lgtm

PR #1087, opened 2023-05-11
aria labels to links fixed the tab index in mobile nav

SamuelmdLow:

The lighthouse report outlined some accessibility issues to look into. I fixed some of them in this branch.

The lighthouse report showed that some links were not labeled in the home page. I labled them. I also found that when in mobile, tabbing in the nav would go to elements outside the page. I fixed that by making the content-wrapper invisible while the nav was open.


SamuelmdLow:
Irrelevant

PR #1091, opened 2023-05-16
Changed image format to webP

SamuelmdLow:

The webP format has higher compression than png. Using it significantly improves the lighthouse report

I also made the column images in the homepage a lower resolution because they were much too high for the small size they were displayed at.

### DIFF
--- a/archive/templates/archive/objects/article_list.html
+++ b/archive/templates/archive/objects/article_list.html
@@ -9,7 +9,7 @@
 	<div class="o-article__left">
 		<a class="o-article__image" href="{% pageurl article %}">
 			{% if article.featured_media.first.image %}
-				{% image article.featured_media.first.image fill-340x238-c100 %}
+				{% image article.featured_media.first.image fill-340x238-c100 format-webp %}
 			{% elif article.featured_media.first.video %}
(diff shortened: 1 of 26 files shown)


psiemens (commented):
Nice PR!

I see that Wagtail allows you to set a [default image conversion mapping](https://docs.wagtail.org/en/v2.8.1/advanced_topics/images/image_file_formats.html#customizing-output-formats) for all images. Would this be better than annotating the templates with `format-webp`, or is there value in only converting specific images to WebP?

```
WAGTAILIMAGES_FORMAT_CONVERSIONS = {
    'bmp': 'jpeg',
    'webp': 'webp',
}
```

psiemens (commented):
Nice PR!

I see that Wagtail allows you to set a [default image conversion mapping](https://docs.wagtail.org/en/v2.8.1/advanced_topics/images/image_file_formats.html#customizing-output-formats) for all images. Would this be better than annotating the templates with `format-webp`, or is there value in only converting specific images to WebP?

```
WAGTAILIMAGES_FORMAT_CONVERSIONS = {
    'bmp': 'jpeg',
    'webp': 'webp',
}
```

SamuelmdLow:
Thank you for looking at this branch! I wasn't aware tat you could change the default image format but I think it would be a better solution because that way we can't forget to add the format-webP to images in future additions to the website. I'll make a branch which does that instead!

PR #1092, opened 2023-05-16
Added icons to images and pull quote blocks

SamuelmdLow:
There was no icons for images and pull quote blocks so I added them


FireBoyAJ24 (approved):
The addition of the icons works.

PR #1093, opened 2023-05-17
Allowed multiple links from different websites

SamuelmdLow:
Before links were displayed in the title in a very ugly way. I fixed that. I also made it so an author can have as many links as they want, not just twitter and facebook. I had it so the website would recognize and give special icons to these platforms:

- tumblr
- instagram
- twitter
- facebook
- youtube
- tiktok
- linkedin
- reddit
- mastodon

Because mastodon is decentralized I had to do an api request to identify a mastodon instance. I'm not aware of a better way to do it


SamuelmdLow:
I will make this better by deciding on link icons when the page is saved, rather than every time it is loaded. I will also add images and videos to the author page as requested by the editors.

PR #1100, opened 2023-05-30
1097 we should add share options to all the articles

### LINKED ISSUE #1097: We should add share options to all the articles
To increase the spread of Ubyssey's articles we can add different share options such as share through the following ways:
1.Facebook
2. Twitter
3. LinkedIn 
4. Print
5. Copying link

image
Is a possible solution

FireBoyAJ24:
Instead of a sidebar. I have chosen to add a discreet horizontal share bar for the main article page, wide page and 2023 magazine page. However, I could not test the guide pages as well as the old magazines pages (prior 2023).

Main article page:
image

Wide article page:
image

Magazine 2023 page:
image

### DIFF
--- a/article/templates/article/article_page.html
+++ b/article/templates/article/article_page.html
@@ -37,6 +37,176 @@
                         </a>        
                     {% endif %} {% endcomment %}
                 </div>
+                <script>
+                    /* Restoring browsering effects */
+*{
+    margin:0;
+    padding: 0;
+    box-sizing: border-box;
+}
+  
+body{
+    height: 100vh;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    background-color: #000;
+}
+  
+.main_box{
+    width: 4em;
+    height: 4em;
+    position: relative;
+}
+  
+#share_button{
+    display: none;
+}
+  
+span,a{
+    position: absolute;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    border-radius: 50%;
+}
+  
+span{
+    width: 4em;
+    height: 4em;
+    top: 50%;
+    left: 50%;
+    transform: translate(-50%,-50%);
+    background-color: #eee;
+    color: #333;
+    font-size: 2em;
+    z-index: 1;
+    cursor: pointer;
+    /* border-radius: 30%; */
+}
+  
+.sm_list{
+    width: 4em;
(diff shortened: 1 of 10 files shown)


SamuelmdLow (changes requested):
Javascript did not work when I tested it. 

The address to share to facebook is missing the argument name "u". It should be  `http://facebook.com/sharer.php?u=' + window.location.href`. 

Also the variable `titleElement` is a list of elements with the tag "title". The innerText of a list returns undefined. titleElement should instead be the first element of this list: `const titleElement = document.getElementsByTagName("title")[0];`

I'm also unsure if these three social medias are actually the best to include. There is a design cost that comes with including unused features. I don't think a lot of university students are posting on linkedin. Even facebook and twitter aren't that widely used for our age group. I think reddit might be more useful to the website - especially because UBC has a pretty active subreddit. It might be useful to ask the editors about this

FireBoyAJ24:
> Javascript did not work when I tested it.
> 
> The address to share to facebook is missing the argument name "u". It should be `http://facebook.com/sharer.php?u=' + window.location.href`.
> 
> Also the variable `titleElement` is a list of elements with the tag "title". The innerText of a list returns undefined. titleElement should instead be the first element of this list: `const titleElement = document.getElementsByTagName("title")[0];`
> 
> I'm also unsure if these three social medias are actually the best to include. There is a design cost that comes with including unused features. I don't think a lot of university students are posting on linkedin. Even facebook and twitter aren't that widely used for our age group. I think reddit might be more useful to the website - especially because UBC has a pretty active subreddit. It might be useful to ask the editors about this

When I began work on the share button feature, I stumbled on a piece of code which allowed us to share articles to different social media apps. The one that was chosen by the previous developer included reddit.

SamuelmdLow (approved):

brittkhat:
I got some weird interactions when I checked the branch in my local dev environment. It produced a text box and the links don't seem to work in my setup. 
image

FireBoyAJ24:
> I got some weird interactions when I checked the branch in my local dev environment. It produced a text box and the links don't seem to work in my setup. image

@brittkhat, have you done gulp build? If we do not do gulp build then the static files will be out of sync with the the static_src. The links will not work as it needs the JS and the textbox needs CSS formatting.

PR #1101, opened 2023-06-05
995 new archive page akshanjay

### LINKED ISSUE #995: New Archive Page

FireBoyAJ24:
1. I worked on adding routable pages to the sections, general archive pages and the video section. 
2. There is still a query selection for the search, order and years filter.
3. I have split the archive filters into a separate html. Whether we wish to keep it as such is up to the reviewer.
4. I have added a link to Ubyssey's Issues page on Issuu on the archive page. It allows users to easily access all the issues Ubyssey has made over the years.

Benefits of the solution:
1. It allows the web developer to add magazines, guides and so on with a greater control over how the section filter will behave.

Drawback with the solution:
1. It is a bit hard to add magazine and guides search capacity with the current setup where the articles within these two categories are hidden. 
2. The way in which I have added the video section (in the same way for future magazine and guide) is a bit clunky in which there is always a continuous check whether the objects are an article or a video. This issue has come because of the resolve to use as few archive templates as possible. This should be further discussed.


SamuelmdLow (commented):
I think everything you added works well. The only thing I would change is that I think it makes more sense for the link to print issues to be labeled "print issues" instead of just "issues"

While reviewing I noticed an issue that pre-exists this branch about tabbing in the modals when in mobile view. I added it as a new issue.

SamuelmdLow (approved):

FireBoyAJ24:
As I have merged the new archive page branch to the magazine branch I believe it is best to merge the pull request of the new magazine filter branch than the new archive page branch.

PR #1110, opened 2023-06-08
1103 have a seperate magazine section for the archive section jay

### LINKED ISSUE #1103: Have a seperate Magazine section for the Archive Section
In the section filters, there have only been 2017, 2018 and 2019 magazine articles that the readers can easily filter for. In the future we want the editorial team to easily add magazines to a separate magazine filter. In addition, we want to reduce the complexity of adding new magazine filters.

A simple solution to generate a new magazine filter:
1. Add a tag to each article which is part of a magazine. 
2. On the Archive page, add a magazine tag query which will accept magazine tags and add them to the magazine filters.

FireBoyAJ24:
Added a separate magazine filter for the archive section. This makes it easy for the editors to add magazine filters but will require that all old articles are tagged using the magazine tag snippet.


SamuelmdLow (changes requested):
`ArchivePage` content panels does not include the default `Page` content panels. This means that you can't set default Page fields like title. If you delete your old archivepage you won't be able to create a new one because required `Page` fields are impossible to fill in admin. 

To fix this you should make it `content_panels = Page.content_panels +[...` so that the default `Page` fields can be filled

archive/templates/archive/objects/archive_filter.html
```diff
@@ -0,0 +1,187 @@
...
+
+    <div class="o-archive__sidebar__filters">
+        <h2>Section</h2>
+        <ul>
+            <li
+                class="o-archive__sidebar__filter{% if not section_slug and video_section == False %} o-archive__sidebar__filter--is-active{% endif %}">
+                <a href="{% pageurl self %}{{ request|query_string }}">All sections</a>
+            </li>
+            {% for section in sections %}
+                {% if section.title != "The Ubyssey Student Marketplace" and "20" not in section.title and  section.title != "Pages" %}
```
SamuelmdLow: This seems like it will easily break. Perhaps instead we should add a checkbox field in sectionPage that, when checked, excludes the section from the Archive

SamuelmdLow (changes requested):
This works. I do think it would be better to rely on the categoriesSnippet instead of new models though. The website already has the magazine articles organized into categories so we could avoid having to relabel years worth of magazine articles if we added the Magazine CategorySnippet as a filter instead.

FireBoyAJ24:
I have incorporated Sam's ideas. Cleaned up the CSS and HTML which includes the current issues that the archive page is facing.

There will be a minimum of 4 category snippets needed.

- 1 for the self

- 3 for the spoofs

In terms of adding the category to articles... It is incredibly simple. We just need to add the above snippets to the sections (For the spoofs it is simple but for the self I had to add a category selection place for this to work)
Up until now, we should not be touching any articles. Effort up to now **MEDIUM**


From what I have found, the old magazine articles are found in the magazine section page. The only issue is that we will have to manually label each and every (Live and public) article to a category snippet.
The number of snippets here can be 2~3. The effort, **HARD (compromises) - IMPOSSIBLE (no compromises).**

SamuelmdLow (approved):

PR #1111, opened 2023-06-09
1096 a dark and light mode for ubysseyca website akshanjay

### LINKED ISSUE #1096: A dark and light mode for Ubyssey.ca website
Having a dark and light mode increases comfort and readability of the article.

What we can expect,
image

FireBoyAJ24:
The dark mode encompasses all the default article and the full width article. The webpage can also remember the user's preference of mode when they set the mode the first time.


SamuelmdLow (changes requested):
- The logo should be a .svg not .png.
- The darkmode logo stretches when in mobile
- The toggle button doesn't fit well with the rest of the page. We use font-awesome so maybe try searching for an icon at fontawesome.com (I'll add the new version of font awesome to your branch)
- Its very jarring when the you switch the modes. Maybe add a transition-duration, and use transition-property so it doesn't affect other properties.
- Most darkmodes don't use such a dark background and bright text colour. For me it makes it difficult to look at. In my redesign I use #031621 as a dark background colour and #D9D9D9 as a dim text colour if you think its a good idea to use those.

ubyssey/static_src/src/images/ubyssey_logo_dark_mode.png
SamuelmdLow, line 1: remove

ubyssey/static_src/src/images/ubyssey_logo_dark_mode.svg
SamuelmdLow, line 1: remove

SamuelmdLow (approved):
We should probable remove those two unused darkmode logos but other than its good!

PR #1112, opened 2023-06-13
1102 modal sam low

### LINKED ISSUE #1102: I think a modal block should exist in base.html
Modals are blocks that appear above the content of a page. Currently our website uses them when selecting options in the archive. The way we do modals now does not work with tabbing. All the content of the page is still  focusable despite not being visible. This disrupts the experience of those using screen readers. The solution I propose is make a block tag for modals `{% block modal %}` in base.html which is above the content block. That way when the modal is opened we can give the content-wrapper div the attribute `aria-hidden="true"` which prevents those blocks from being focused on. My mockups propose using modals and this solution would make it easier to add them and reuse code and maintain the same style across  them.

Another issue I noticed is that the close button is not focusable. It should should be surrounded by an anchor tag so that the user can focus on it and press enter to close. It would also be nice if the modal closed with the esc button.

SamuelmdLow:

Previously all modals were made independently. This branch unifies the modals. The branch also separates the modals from the content so that the content of the page can be given the attribute "inert" so that the user cannot interact with it (focusing, clicking links, etc).

Modals are now made like this

```

<div class="open-modal">
	<a href="#">OPEN MODAL BUTTON</a>

        // MODAL DIV
	<div class="add-to-modal ">
		<a class="close-modal" href="#">CLOSE MODAL BUTTON</a>
	</div>

</div>

```

The element with class "add-to-modal" is moved by javascript into the modal which is outside of "content-wrapper". In its place is left an element like this `<div class="openModal" modal="0"></div>`.  The attribute `modal` is the index of the element in the modal. Clicking the open modal button would open the modal and show the element at that index.


SamuelmdLow:
Merged into other branches

PR #1114, opened 2023-06-13
Added the redirect from blog to humour

### LINKED ISSUE #1113: make blog articles redirect to humour
Editors want to rename the blog section "humour". Changing the slug of this section would cause link rot for old blog articles. To stop that we should redirect `/blog/... ` to `/humour/...`

SamuelmdLow:

Editors want to rename blog to humour. Changing the slug would cause link rot for blog articles. This redirect would prevent the link rot


SamuelmdLow:
I've realized that once slugs are changed old artices still have the unupdated slug saved and our archive thing fails to retreive them. I'm working to fix this by changing the way we retrieve articles to looking for child pages

PR #1116, opened 2023-06-14
1113 redirect blog to humour samlow

SamuelmdLow:

Editors wish to rename the blog section to humour. Changing the slug would cause linkrot and also prevent all previous articles from being retrieved in the section page.

To prevent link rot I created a redirect  for requests to /blog/... to /humour/...

To fix the way sectionpage retrieves articlesI had it relied on the page hierarchy (parent/child pages) rather than the field "current_section" which will become out of date when the slug is changed


SamuelmdLow:
When we push this to production we have to very quickly change the slug in admin from blog to humour. In the time that we don't, all blog articles will be unreadable as the site will redirect the user to a /humour/ which does not yet exist

brittkhat (approved):
lgtm

PR #1118, opened 2023-06-15
Adding subfolders to the wagtail admin

### LINKED ISSUE #1115: Adding subfolders in the section's folder
Issue: Currently, on the Ubyssey's directory there are a few spoof sections which are technically a sub section of the blog section. As the spoof sections are in the sections directory the archive filters considers these sections as any other section. This has caused unnecessary disorganization for the archive filters and can possibly turn into a problem if the method of retrieving these sections are used in any other part of the website.

Proposal: We should add a subfolder so that we can move these spoof sections into the blog section. This makes it easier for the editors to be organized and find the articles easily.

FireBoyAJ24:
Added a base sub folders for the wagtail admin. The subfolders is similar to the section folder but has no category snippets in place. The sub folders are intended to store all the spoof articles and the sex issue in the blog folder.


SamuelmdLow (commented):
This would solve the issue you wrote about. I'm a little confused though because we could also just add the articles to a category and move them back to one of the main sections without introducing subsections.

FireBoyAJ24:
> This would solve the issue you wrote about. I'm a little confused though because we could also just add the articles to a category and move them back to one of the main sections without introducing subsections. 

The main purpose of this issue is to provide the editors with an easier time finding the spoof articles in the specific spoof folder than in the blog section.

FireBoyAJ24:
This pull request is not required as we are not using magazine snippets.

PR #1119, opened 2023-06-15
Redone authors page

### LINKED ISSUE #996: New Authors Page
This text is largely copied from #995. Archive and Author pages in any case will have similar designs. Fundamentally the divergence between what we have and what we need is a place to show off _all_ student created content, not just articles.

Author Pages have usually been thought of as like Archive Pages that further filter the queries by the particular author. This may or may not be the best design decision but it's why these issues are so similar.

Author Pages have work to do that is not shared by the Archive Page however. They've traditionally been thought of as a sort of portfolio for particular students' work. This raises many of the same problems that hosting old student articles for a long time also raises. With frequent organizational turnover and needing to constantly update and maintain the site, I don't know if it's reasonable to make the promise to all future student writers that their work will never disappear.

---

SamuelmdLow:
- Authors can have as many links as they want
- Includes a link to their rss feed (The rss feed is added in the rss branch - not this one)
- Photos and videos are shown on authors page
- Authors can pin articles to the top of their author's page
- The authors page is redesigned


SamuelmdLow:
Inside redesign

PR #1121, opened 2023-06-16
Added sidebar category block

### LINKED ISSUE #1120: Changing obituaries section to obituary snippet
"Is it possible to get the “obituaries” section on the front page tied to the snippet/category called “obituaries” instead of the subsection? Just so everything can be nested under features since that's the assigning section." - From the editorial team.

The obituary articles presentation on the homepage must behave similarly as before. Since, we have still not released the new homepage design further adjustments maybe needed.

FireBoyAJ24:
To be able to move the obituary to the feature section, I had to add a sidebar category block which has the same html template as the sidebar section block. However, the sidebar category articles received the category articles instead of the section articles.


SamuelmdLow (changes requested):
I don't think the sidebar category block needs its own template since `sidebar_category_block.html` is exactly the same as `sidebar_section_block.html`

PR #1124, opened 2023-06-27
Redesign!!!

### LINKED ISSUE #995: New Archive Page

### LINKED ISSUE #1041: Feature: suggested articles

### LINKED ISSUE #1102: I think a modal block should exist in base.html

### LINKED ISSUE #1103: Have a seperate Magazine section for the Archive Section

### LINKED ISSUE #1108: audio quote block
Editors suggested we add an audio quote block like in this article

https://www.nytimes.com/interactive/2023/05/11/opinion/connie-chung-named-after.html?unlocked_article_code=_ciEExomEiuJjAF5TB-tMVjpvEIfyuiPewMW-ITFh6aHNGlFcQWffazK77_YvlIUoZ7w_d1szZdhxa6oQmfNAlRGTykBLjsjXin-FOnic5mdMhJE65ynL6W97ac3bL8b8Q3sqnbZgteXxJuXUF1PqG4w12F1Oah94J8WDAth7ZTRcKqG0wk0igMzd1kypVrg4t972cwAD-0cZAbsQuJmNsZ7ze2m5D2zzIpMZ3U0ZvbQ83vpXDqhdX3PELwPAVGpFTmNXtzhZFwYbrT_mjgytb-8oMTRDaNNIo84ILfGqxBqBqL3m1h9MHMHu-fCdS3qdwROI8dod6jxmwFKWhQlIvpm6PA48yFZf7VfcJfybNX-ig&amp;smid=url-share

### LINKED ISSUE #1120: Changing obituaries section to obituary snippet

### LINKED ISSUE #1154: Mobile Leaderboard Ads are not present
Mobile leaderboard ads are not presented on the home page and in between articles. The issue may come down to an error in the Adslot model or in the google ads API.

### LINKED ISSUE #1155: The author's bio is in the pages section
## What behavior were you expecting?

image

## What actually happened?
Additional consideration to the display in the pages section was not considered

## Steps to reproduce
Go to any page that is within the pages section. For example, volunteer, advertise, about, ext.

### LINKED ISSUE #1160: Rework the One of Video StreamBlock
## What Happened?
As I was going through the static files to understand whether they where used or not, I found a Youtube.jsx file in the modules folder. It is currently being used in our site. However, it is broken as I cannot start the video after clicking the button. 

image

The method at which any embeds occur is through the use of raw html stream-blocks compared to the use of the embed in richtext and the one off video stream-block.

## What Should be Done?
I am unsure of the extent at which this block is being used. Should we rework the stream-block and keep it or completely remove it? The only thing I am clear on doing is to remove the Youtube.jsx and unwind the spaghetti code.

### LINKED ISSUE #1163: Add a Redact Option in the Rich Text Editor
## What behavior were you expecting?
Using a span block the writers where able to redact phrases from the text. However, we did not account for the needed colour inversion of the redact in the dark mode. The resulting error can be seen in the dark mode image.
image
image

## What actually happened?
The writers where able to add a redact by adding a raw HTML block and then adding a background-colour: black in the inline style section. 

## What can we do?
In the future the addition of a highlighting feature in the rich text editor would be preferable. The redact feature will have to be separated as that is not a highlighting feature. We will need to add a special html block in the wagtail web hooks and a rich text editor button for redact, and in the future for the highlighting feature.

SamuelmdLow:
image

image

image


archive/models.py
```diff
@@ -39,37 +42,50 @@ def __parse_int_or_none(self, maybe_int):
...
+                social_cover_image
 
-        # Getting information from the HTTP request
+                section_name
+                issue
+                section_image
+                
+    @route(r'^year/?$')
+    ..... (includes the section and year route)/search=q or /search=".........."
+    """
```
FireBoyAJ24: These comments are not needed.

archive/templates/archive/archive_page.html
```diff
@@ -12,9 +13,15 @@
         <div class="c-page__header">
             <h1 class="c-page__heading">Search the archive</h1>
```
FireBoyAJ24: Just need to capitalize the archive name. Do we need to have the heading very large.

archive/templatetags/archive_tags.py
```diff
@@ -17,6 +19,18 @@ def modify_query_string(context, field, value):
...
+def query_string(request):
+    fullurl = request.get_full_path()
+
+    fullurllst = fullurl.split("?")
+    if "?" in fullurl:
+        parameters = "?" + fullurllst[1]
+    else:
+        parameters = ""
+
+    return parameters
```
FireBoyAJ24, line 33: Could we have used {{ request.GET }} instead of making a separate tag.

article/templates/article/objects/blog_column.html
```diff
@@ -8,7 +10,7 @@
 {% load wagtailcore_tags %}
 {% load wagtailimages_tags %}
 {% load video_filters %}
-<article class="o-article o-article--column{% if padded %} o-article--padded{% endif %}">
+<article class="o-article o-article--column{% if padded %} o-article--padded{% endif %}" style="background-color: {{article|get_colour}}">
```
FireBoyAJ24, line 13: Is there meant to be a CSS style in the article tag? Can the style be placed in a class, or is it too much?

FireBoyAJ24 (dismissed):
1. There is a small bug where I see show more when the homepage page states that I have reached the end. In addition, when I search using the search bar in the section page, if there is no results it says, You reached the end.

2. The loading sign is oddly placed in the mobile screen.

3. The debug system sends a complaint with the function, get_paginated_articles()
to replicate the error, http://localhost:8000/archive/?q=Type-anything-here.

4. If the author's bio is long, the container cuts into the nav bar.

5. Do you want to keep the formatting the same for the section page and the archive page? For example, the archive page will have an infinite feed or the section page will not have an infinite feed.

SamuelmdLow:
To do: Add dark mode

FireBoyAJ24 (approved):
Asa beta I accept the redesign. We may still want to discuss about the suggested articles style as the sidebar suggested articles may feel a bit jarring.

brittkhat:
lgtm

PR #1134, opened 2023-07-04
Changed the google tag again

SamuelmdLow:
Thats the id google analytics says we should use now. I'm not sure why it changed


brittkhat (approved):
lgtm

PR #1150, opened 2023-07-13
Redid merges

### LINKED ISSUE #1151: General Visual Errors in the Author's Bio
- When on article pages, the top of the header is cut off so there's no way to get back to the home page
- Photo captions should be grey, right now they're black
- Author links on articles still leading back to the article page
- The social media links on article pages go further left than the article itself - can the articles be slightly wider to match that
- There's a lot of negative space around all author names right now, I've messaged Jay about this directly too but can that be decreased + authors be left-aligned, particularly if there's no photo?

SamuelmdLow:
Last merge reverted a lot of comitts. I redid the merges


article/models.py
```diff
@@ -178,18 +178,6 @@ class ArticleAuthorsOrderable(Orderable):
...
-                            ("news producer", "News Producer"),
-                            ("features editor", "Features Editor"),
-                            ("humour editor", "Humour Editor"),
-                            ("culture editor", "Culture Editor"),
-                            ("opinion editor", "Opinion Editor"),
-                            ("video editor", "Video Editor"),
-                            ("photo editor", "Photo Editor"),
-                            ("visuals editor", "Visuals Editor"),
-                            ("sports editor", "Sports Editor"),
-                            ("science editor", "Science Editor"),
```
FireBoyAJ24, line 192: These must be added as Anabella has requested them

article/models.py
```diff
@@ -991,7 +980,7 @@ def get_authors_with_urls(self) -> str:
     authors_with_urls = property(fget=get_authors_with_urls)
 
     def get_authors_in_order(self):
-        AUTHOR_TYPES = ["author", "photographer", "illustrator", "videographer", "coordinating editor", "news editor", "news producer", "features editor", "humour editor", "culture editor", "opinion editor", "video editor", "photo editor", "visuals editor", "sports editor", "science editor"]
+        AUTHOR_TYPES = ["author", "photographer", "illustrator", "videographer"]
         authors = self.article_authors.all()
```
FireBoyAJ24, line 985: These must be added as Anabella has requested them

article/templates/article/article_page_guide_2020.html
```diff
@@ -48,6 +48,9 @@
                                 {{ article.template_fields.subheading }}
                             </p>
                         {% endif %}
+                        <div class="c-article__byline__meta">
+                            {% include 'article/objects/darkmode_button.html'%}
+                        </div>
```
FireBoyAJ24: These must be removed

article/templates/article/article_page_guide_2022.html
```diff
@@ -72,6 +72,9 @@
                         {{ featured_image_object.credit }}
                     {% endif %}
                 </div> {% endcomment %}
+                <div class="c-article__byline__meta">
+                    {% include 'article/objects/darkmode_button.html'%}
+                </div>
```
FireBoyAJ24: These must be removed

ubyssey/static_src/src/styles/modules/_darkmode.scss
```diff
@@ -1,16 +1,18 @@
 @import 'variables';
 
-.c-article_dark_mode {
-  float: right;
-}
-
```
FireBoyAJ24, line 6: Why is this getting deleted? The class is ensuring that the dark mode toggle is to the right of the share button's bar

FireBoyAJ24 (changes requested):

ubyssey/static_src/src/styles/modules/_darkmode.scss
SamuelmdLow: I didn't think it looked very good like that

article/models.py
SamuelmdLow: I think its more elegant if we have one of the options to be showing the author's organization role. I've added that. If we think we need to have every organization role in the selector then we can add that

ubyssey/static_src/src/styles/modules/_darkmode.scss
FireBoyAJ24: Combining the share bar and the dark mode button makes it look a bit untidy. That is why I placed it to the right of the line.

SamuelmdLow: Lets ask editors what they prefer

brittkhat (approved):
lgtm

PR #1152, opened 2023-07-14
Rss fix samlow

SamuelmdLow:
Once it was on the live website I noticed some pretty major issues with the rss feed which I have fixed in this branch

- It displayed the oldest articles instead of newest
- The default django rss feed returns the wrong content type for the httpresponse for some reason. I altered the class to make it correct
- items in the feed were not provding the full url so rss readers didn't know to send the user to ubyssey.ca

I've also added featured media images to the item description for quality of ife


brittkhat (approved):
lgtm

PR #1156, opened 2023-07-19
Replace Dispatch user model with custom model

psiemens:

The Django/Wagtail project still imports code from Dispatch. Most of these imports come from dead code (i.e. very few live pages still depend on Dispatch), but the Wagtail admin dashboard still uses the `dispatch.User` model for user authentication. This requires us to keep Dispatch as a dependency, not only in `requirements.txt` but also as an installed Django app.

Removing `dispatch.User` from this site gets us one step closer to removing Dispatch altogether.

This PR adds the `users.User` model, a custom user model to replace `dispatch.User`. Rather than reverting to the Django default user model, I opted for a custom model to preserve the current email-based login system.

These are the steps to migrate to the new user model:

1. Manually add `users.0001_initial_1` to the `django_migrations` table. This prevents this error:
```
django.db.migrations.exceptions.InconsistentMigrationHistory: Migration authtoken.0001_initial is applied before its dependency users.0001_initial on database 'default'.
```
2. Duplicate the `dispatch_user` table, name the new table`users_user`
3. Run `python manage.py migrate`. This run apply the `users.0002_initial_2` migration.
4. Optional: populate first_name, last_name fields from person table


keeganland (commented):
We should be certain whether the "email" field in user model is validated as containing an email formatted string.

psiemens:
@keeganland Good catch. I updated the email field to Django's built-in [EmailField](https://docs.djangoproject.com/en/4.2/ref/models/fields/#emailfield) (https://github.com/ubyssey/ubyssey.ca/pull/1156/commits/a4f2b516b7a6ca19260da8d7c7e87f381a590625), which is a small wrapper that adds email validation to CharField.

Wagtail already rejects invalid emails in its user forms, but now we have validation at the model level as well.

Screenshot 2023-07-24 at 3 17 13 PM
<br/>
Screenshot 2023-07-24 at 3 14 05 PM

brittkhat (approved):
lgtm

PR #1165, opened 2023-08-01
Added redact and fixed the right and center bugs

### LINKED ISSUE #1163: Add a Redact Option in the Rich Text Editor

FireBoyAJ24:

#1163 

I have added the redact feature to the rich text editor and fixed the issue where the right and center text are not displayed when we select them.


SamuelmdLow:
The styling in the tag caused an error where the redact, center, and right aligns would disappear when the article was saved. I removed it and moved the styling to the stylesheet.

SamuelmdLow (approved):

PR #1220, opened 2023-08-17
Creates  "More" drop down in the main nav

SamuelmdLow:
image

When we merge this we must move items from the left and right header menu into a new flat menu called "dropdown_menu"


brittkhat (approved):
lgtm

PR #1221, opened 2023-08-17
Fix special article page

SamuelmdLow:
The footer would raise to the top of the article and block other elements from being clicked on or highlight. I fixed this by making the container a flex box


brittkhat (approved):
lgtm

PR #1224, opened 2023-08-19
Fixed spacing in articles

SamuelmdLow:
I removed a `margin: auto` which overwrote the margins of certain elements inside `article-content`


brittkhat (approved):
lgtm

PR #1225, opened 2023-08-19
Fixed some issues with section/category info

SamuelmdLow:
I removed empty divs and weird margins. I also removed featured categories from categories


brittkhat (approved):
lgtm

PR #1251, opened 2023-08-24
Made fields in the promote tab actually work!

### LINKED ISSUE #1250: We do not add meta keywords
We have an unused field for seo keywords but they are not added to the head

This is bad for our SEO

SamuelmdLow:
Meta description, Title tag, and SEO keywords fields did not work. I edited the meta_tags template to include them.

meta keywords now also defaults to the tags field when seo keyworks is blank because writers have apperently been using it for this purpose


brittkhat (approved):
lgtm

PR #1253, opened 2023-08-24
Fixed quote block issue

### LINKED ISSUE #1252: Quote block says "Listen to" even when there is no audio
It should not say that

SamuelmdLow:
The quote block was saying "Listen to [source]" even when there was no audio. I replaced "Listen to" to a dash


brittkhat (approved):
lgtm

PR #1255, opened 2023-08-25
1254 special landing page coverstory samlow

### LINKED ISSUE #1254: Allow special landing page as coverstory
To feature the guide and magazine we should allow special landing pages as the coverstory. Currently these pages do not have featured media or lede so we will have to add these fields as well

SamuelmdLow:
I gave the special landing page a featured_media field and allowed other types of pages to be the coverstory


brittkhat (approved):
lgtm

PR #1257, opened 2023-08-25
Fixed order filter

### LINKED ISSUE #1256: Order filter doesn't close when clicked off
The options of the order filter should dissapear when its not focused on

SamuelmdLow:
The 'order' filter on the archive page would not dissapear when it was unfocused on. I fixed this by moving the mechanism out of javascript and into scss. I also added the label for the order filter to the authors page


brittkhat (approved):
lgtm

PR #1335, opened 2023-10-26
To reduce the time to order the pages, I restricted the pool to only …

SamuelmdLow:
To reduce the time to order the pages, I restricted the pool to only articles that were published within two months.

is this good practice?

### DIFF
--- a/section/models.py
+++ b/section/models.py
@@ -305,13 +305,10 @@ class SectionPage(RoutablePageMixin, SectionablePage):
         return context
     
     def get_section_articles(self, order='-explicit_published_at') -> QuerySet:
-        # Optimization: limit queryset to articles published in the last 60 days to shorten query time
-        recent_articles_delta = datetime.datetime.now()-datetime.timedelta(days=60)
 
         section_articles = ArticlePage.objects \
             .live() \
             .child_of(self) \
-            .filter(explicit_published_at__gt=recent_articles_delta) \
             .order_by(order)
         
         return section_articles
--- a/section/models.py
+++ b/section/models.py
@@ -307,7 +307,14 @@ class SectionPage(RoutablePageMixin, SectionablePage):
         # return ArticlePage.objects.from_section(section_root=self)
         # section_articles = ArticlePage.objects.live().public().filter(current_section=self.slug).order_by(order)
         #section_articles = ArticlePage.objects.live().public().descendant_of(self).order_by(order)
-        section_articles = ArticlePage.objects.live().child_of(self).filter(explicit_published_at__gt=datetime.datetime.now()-datetime.timedelta(days=60)).order_by(order)
+        # Optimization: limit queryset to articles published in the last 60 days to shorten query time
+        recent_articles_delta = datetime.datetime.now()-datetime.timedelta(days=60)
+
+        section_articles = ArticlePage.objects
+            .live() \
+            .child_of(self) \
+            .filter(explicit_published_at__gt=recent_articles_delta) \
+            .order_by(order)
         
         return section_articles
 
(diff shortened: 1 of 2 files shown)


section/models.py
```diff
@@ -303,9 +303,12 @@ def get_context(self, request, *args, **kwargs):
         return context
     
     def get_section_articles(self, order='-explicit_published_at') -> QuerySet:
+        import datetime
```
psiemens: Best to put this with the rest of the imports at the top of the file

section/models.py
```diff
@@ -303,9 +303,12 @@ def get_context(self, request, *args, **kwargs):
...
         # return ArticlePage.objects.from_section(section_root=self)
         # section_articles = ArticlePage.objects.live().public().filter(current_section=self.slug).order_by(order)
-        section_articles = ArticlePage.objects.live().public().descendant_of(self).order_by(order)
+        #section_articles = ArticlePage.objects.live().public().descendant_of(self).order_by(order)
```
psiemens: To keep the code clean and avoid confusion, I'd recommend deleting these commented lines. We can always look back in the git history to see previous versions of the code

section/models.py
```diff
@@ -303,9 +303,12 @@ def get_context(self, request, *args, **kwargs):
...
+        section_articles = ArticlePage.objects.live().child_of(self).filter(explicit_published_at__gt=datetime.datetime.now()-datetime.timedelta(days=60)).order_by(order)
+
```
psiemens: I noticed that this queryset no longer includes the `.public()` selector, which AFAIK was there to prevent private pages from showing up on the homepage.

When profiling the site I discovered that the `.public()` selector slows queries down quite a bit, I think because Wagtail doesn't create a database index for public/private pages (`.live()`, on the other hand, is indexed). I recommend adding `.public()` back and remeasuring the query times. It could be that most of your speed up is coming from the removal of this call rather than the new filter. If that's the case, unfortunately I don't think it's worth it to add the complexity of the 60 day filter.

I've been wondering if we could get away without the `.public()` selector if private pages aren't being used, but wanted to play it safe and explore that optimization later.

section/models.py
```diff
@@ -303,9 +303,12 @@ def get_context(self, request, *args, **kwargs):
         return context
     
     def get_section_articles(self, order='-explicit_published_at') -> QuerySet:
+        import datetime
         # return ArticlePage.objects.from_section(section_root=self)
         # section_articles = ArticlePage.objects.live().public().filter(current_section=self.slug).order_by(order)
-        section_articles = ArticlePage.objects.live().public().descendant_of(self).order_by(order)
+        #section_articles = ArticlePage.objects.live().public().descendant_of(self).order_by(order)
+        section_articles = ArticlePage.objects.live().child_of(self).filter(explicit_published_at__gt=datetime.datetime.now()-datetime.timedelta(days=60)).order_by(order)
```
psiemens: Some suggestions to make the code more readable for future maintainers:

- Move time delta to a variable with a descriptive name
- Comment explaining the rationale behind this optimization
- Split queryset function calls on to separate lines for readability 

```suggestion
        # Optimization: limit queryset to articles published in the last 60 days to shorten query time
        recent_articles_delta = datetime.datetime.now()-datetime.timedelta(days=60)

        section_articles = ArticlePage.objects
            .live() \
            .child_of(self) \
            .filter(explicit_published_at__gt=recent_articles_delta) \
            .order_by(order)
```

psiemens (commented):
This is a really great idea!

section/models.py
SamuelmdLow: I think you are right. It seemed to reduce queries (269->260) but I think the load time was actualy increased a bit.

We don't have any pages deeper than the section pages so I don't beleive we need `.public()`

PR #1337, opened 2023-10-27
Add Wagtail image renditions cache

psiemens:
I noticed that we weren't making use of the `renditions` cache, so I added it in. On my local installation, this further reduces the homepage query count from 400 down to 283.


SamuelmdLow:
👍

PR #1369, opened 2023-12-06
Fix Wagtail 3 upgrade

psiemens:

The last deployment failed due to a few errors in the Wagtail 3 upgrade PR. This PR attempts to fix them!

- Bump production runtime environment from Python 3.7 to Python 3.9
  - One of the new package versions (`django-debug-toolbar`) is only compatible with Python 3.8 and above. We could downgrade `django-debug-toolbar` instead, but it's past time to upgrade Python anyways and 3.9 is very backwards compatible with 3.7.
- Fix diverging migration histories in the `home` and `specialfeaturelanding` apps.


brittkhat (approved):
lgtm

PR #1379, opened 2023-12-12
Add audio block

### LINKED ISSUE #1346: Create audio block
We have an audio option for pull quotes but we have been requested to include a normal audio block, probably for longer audio

This block would include the fields
audio: mp3 file
caption: text

SamuelmdLow:
Audio control with caption text

I didn't do anything interesting with the styling. I think we should look into using other audio control projects instead of making our own. We can do that later though.

This has migrations for the audio block


brittkhat (approved):
lgtm

PR #1381, opened 2023-12-15
Add empty article template

### LINKED ISSUE #1352: Create new blank template
I recently had to fix [this really amazing article](https://www.ubyssey.ca/science/where-we-draw-the-line/) which had its own unique styling and layout. The problem was that the custom html template referenced a thing we no longer have in the backend (dispatch template tags).

To allow for articles like this to be easier to make and less likely to break due to backend changes, I propose we create an "empty" template which only displays the content blocks with no other elements (title, author byline, suggested articles, ads, etc). Since these are for one off articles, any extra layout elements can be included through the raw_html content block.

We should also make a standard custom css file that goes with this template so that people using it can easily add to, remove from, and make changes to the styling.

SamuelmdLow:
Unlike the print issues, digital articles mostly all have the same styling. Not creative. Very boring. Sad. Here is an exception where styling and templating was actually designed to support the article's subject: http://ubyssey.ca/science/where-we-draw-the-line Cool! Amazing! An inspiration! But why are articles like it so rare? Because the barrier to custom styling is incredibly high. To accomplish the custom design in "Where we draw the line" the author had to write a whole custom css styling and custom html template FROM SCRATCH and then upload the template as a document to the cms and commit the styling to this git repository. In the interest of removing barriers to creativity, this pull request solves some issues with this

1. This pr adds an empty template option. This removes everything except for the content blocks. That means that changing the layout of an article does not require writing a template from scratch. In addition to being easier, this also solves a problem I had to deal with recently where the "Where we draw the line" custom templates failed as it included a template tag that no longer exists. The same could occur if it included fields that have been changed/removed. Now that we can just build off an empty template, custom layouts will not be entangled with the backend in this way.
2. This pr makes the custom css styling field actually work now. In the "Custom Frontend" tab in the cms for an articlepage, there is a field called "Custom CSS" which for some reason just did  **n o t h i n g**. This pr makes it do something. Specifically it links the style sheets uploaded to the field. This means custom styling does not require interacting with the git repositry. It can just be uploaded as a document to the cms.

What does this pr not solve?

1. It is now easier to do custom styling but that does not mean it is easy. Most editors and writers do not know how to write css or html. If us webdevelopers are interested in allowing more creativity on the website, we must become instruments for this creativity. We need to become more involved with the website's contents by offering to use our webdesign knowledge to supplement it.
2. This pr does not change the expectations of writers and editors. After years of custom styling almost being impossible and the webdev team's detachment from the journalism office, we really need to advocate for our involvement with content to change expectations. Everyone needs to know that we want to work closer with writers so that more beautiful articles like "Where we draw the line" can get published on the site.

I feel strongly about this because I love and care a lot about the web. I see enormous potential in what websites can be. But when I look around and see whats actually there, overwhelming I find that pages on the web are less creative, less artistic, and less fun than their print equivalents. I hope ubyssey.ca can become part of a movement towards more creativity in design on the web!


SamuelmdLow:
Heres a screenshot of a recreation I did of "Where we draw the line" using the custom css field functionality and empty template added in this pr
image

brittkhat (approved):
lgtm

PR #1452, opened 2024-05-09
Galllery block creation

### LINKED ISSUE #1117: Gallery Block does not work
The Gallery block in the article content streamfield has a button to open the gallery. Clicking this button does nothing

Nishim12:
I added a gallery block

Resolved issue number #1117
I added a gallery block and tested its working
I added a gallery block


SamuelmdLow:
Here are the changes I made
* Templates don't need the html, head, body, or meta tags. Those are only needed once in a page so I removed them and moved the bootstrap stylesheet and script to the head of the page
* I said to make an image, caption, and credit fields in a structblock but I realized it would be better to use the already created Imageblock to put inside the listblock so I changed it to that
* I changed some styling to make it more aligned with the mockup and also revert back things that were altered by the bootstrap stylesheet
* I added an id field to the galleryblock so that multiple galleries can exist

SamuelmdLow (approved):
Here are the changes I made

* Templates don't need the html, head, body, or meta tags. Those are only needed once in a page so I removed them and moved the bootstrap stylesheet and script to the head of the page
* I said to make an image, caption, and credit fields in a structblock but I realized it would be better to use the already created Imageblock to put inside the listblock so I changed it to that
* I changed some styling to make it more aligned with the mockup and also revert back things that were altered by the bootstrap stylesheet
* I added an id field to the galleryblock so that multiple galleries can exist
* I added the gallerysnippet back so that we wouldn't lose connection to what old article used what snippet. That way we can do a migration in the future to make the gallerysnippet using article use our new gallery block

PR #1529, opened 2024-08-21
1528 author gets listed twice when included for two roles

### LINKED ISSUE #1528: Author gets listed twice when included for two roles
Sometimes an author has multiple roles such as writer and illustrator in an article like in https://ubyssey.ca/features/how-ubc-publications-navigate-the-online-news-act/

This means they get listed twice in the authors field. The function `get_authors_with_urls` in articlepage doesn't account for duplicates so it
lists the name twice. The authors block at the bottom of the page also doesn't account for duplicates so that author gets listed twice as well

Nishim12:
It filters duplicate author links and author cards


article/templates/article/objects/author_card.html
SamuelmdLow, line 1: I'd prefer the logic for this to be server side. I think it would be neater and probably quicker

article/models.py
SamuelmdLow, line 1: It doesn't really matter but I think it would make more sense to use author.id rather than full name. There probably won't ever be two different authors with the exact same name but its probably better practice to account for the possibility

SamuelmdLow (changes requested):
Could you identify unique authors by id rather than full name and move the logic for the author card server side instead of in javascript

PR #1578, opened 2025-01-16
Optimize performance on mobile

Nishim12:
It improves the performance of our website on mobile

The way I improved performance is by giving options for multiple size images so that we load the image based on the size of the device so we dont use a large image but rather a smaller image thus reducing our time to load the image as checked by lighthouse reports.


config/settings/base.py
```diff
@@ -293,6 +293,7 @@
     'canonical_domain.middleware.CanonicalDomainMiddleware',
     'whitenoise.middleware.WhiteNoiseMiddleware',
     'django.middleware.gzip.GZipMiddleware',
+    'wagtailcache.cache.FetchFromCacheMiddleware',
```
SamuelmdLow: FetchFromCacheMiddleware is supposed to be last for some reason https://docs.coderedcorp.com/wagtail-cache/getting_started/install.html

We already install it last

SamuelmdLow (approved):

PR #1611, opened 2025-03-17
Update codebase for Linux deployment

psiemens:
This PR adds code to deploy the website to a Linux VM. See the documentation in the [deployment.md](https://github.com/ubyssey/ubyssey.ca/blob/staging-deploy/deployment.md) readme for more info!

Please let me know if there is anything I missed in the readme!


docker-compose.yml
```diff
@@ -0,0 +1,80 @@
...
+      - GS_SECRET_ACCESS_KEY
+      - GOOGLE_APPLICATION_CREDENTIALS
+    expose:
+      - 8000
+    depends_on:
+      - mysql
+      - cache
+
+  nginx:
+    image: nginx:latest
```
Nishim12: Would it better if we use a specific image tag version for this nginx image just so that if they update the image in future we dont see unexpected errors?

docker-compose.yml
```diff
@@ -0,0 +1,80 @@
...
+    restart: always
+    volumes:
+      - ./nginx/:/etc/nginx/:ro
+      - ./certbot/www/:/var/www/certbot/:ro
+      - ./certbot/letsencrypt-etc/:/etc/letsencrypt/:ro
+    depends_on:
+      - django
+  
+  certbot:
+    image: certbot/certbot:latest
```
Nishim12: Would it better if we use a specific image tag version for this nginx image just so that if they update the image in future we dont see unexpected errors?

.github/workflows/build.yml
```diff
@@ -0,0 +1,56 @@
...
+            type=sha
+
+      - name: Build and push Docker image
+        id: push
+        uses: docker/build-push-action@v6
+        with:
+          context: .
+          push: true
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
```
Nishim12, line 56: I can add it later after the current code is merged to add caching just to speed up build times for docker images.
     
```yaml
- name: Build and push Docker image
  id: push
  uses: docker/build-push-action@v6
  with:
    context: .
    push: true
    tags: ${{ steps.meta.outputs.tags }}
    labels: ${{ steps.meta.outputs.labels }}
    cache-from: type=gha
    cache-to: type=gha,mode=max
```          
          
 Reference from https://docs.docker.com/build/ci/github-actions/cache/

.github/workflows/deploy-staging.yml
```diff
@@ -0,0 +1,87 @@
...
+        with:
+          python-version: '3.9'
+
+      - name: Install Python dependencies
+        run: python -m pip install -r requirements.txt
+
+      - name: Collect Django static files
+        run: DJANGO_SETTINGS_MODULE=config.settings.base python manage.py collectstatic --noinput
+
+      - uses: 'google-github-actions/auth@v2'
```
Nishim12: A very minor comment but would be great if you can please add a `-name` maybe to explain that this step essentially authenticates to the GCP

Nishim12 (approved):
Thanks a lot peter for this help to migrate from App Engine to Compute Engine. It will really help us. Just have a few very minor comments.

.github/workflows/deploy-staging.yml
```diff
@@ -0,0 +1,87 @@
...
+        with:
+          workload_identity_provider: 'projects/1012602718138/locations/global/workloadIdentityPools/ubyssey-wif-pool-staging/providers/ubyssey-oidc-github-staging'
+          service_account: 'deployment@ubyssey-staging.iam.gserviceaccount.com'
+
+      - uses: 'google-github-actions/upload-cloud-storage@v2'
```
Nishim12: Same comment here also please

docker-compose.yml
```diff
@@ -0,0 +1,80 @@
...
+  mysql:
+    image: mysql:8.0
+    environment:
+      MYSQL_DATABASE: ubyssey
+      MYSQL_USER: ubyssey
+      MYSQL_PASSWORD_FILE: /run/secrets/SQL_PASSWORD
+      MYSQL_ROOT_PASSWORD_FILE: /run/secrets/SQL_PASSWORD
+    secrets:
+       - SQL_PASSWORD
+    volumes:
```
Nishim12, line 13: Should we specify the option to restart all the containers on failure like
```yaml
deploy:
  restart_policy:
    condition: on-failure
    delay: 5s
    max_attempts: 3
    window: 120s
```
 just so that if containers crash they are automatically restarted in production for all containers?
 https://docs.docker.com/reference/compose-file/deploy/

docker-compose.yml
psiemens: Yes, that's a good idea!

docker-compose.yml
psiemens: Yep, same here.

.github/workflows/build.yml
psiemens: Awesome. Yes, let's try adding this after we confirm that the current setup is working.

.github/workflows/deploy-staging.yml
psiemens: This is a great idea, thanks!

docker-compose.yml
psiemens: Yes, this is also a good idea. I'll add restart policies to our containers

psiemens:
Great review @Nishim12, thanks!

PR #1633, opened 2025-05-09
Primary tag as dropdown

SamuelmdLow:
Made primary tags as a dropdown

It is initialized with every tag ever as an option. (**This is very bad and annoying and I hate it.**  But I can't figure out how to pass page specific information into the widget. Ideally it is initialized with the choices as only the tags in the tags field)

then after rendering it removes everything that isn't in the tags field with javascript

used callbacks provided by https://github.com/wagtail/wagtail/blob/main/wagtail/admin/static_src/wagtailadmin/js/vendor/tag-it.js to add new tags to the primary tags field dropdown as they are added. Also removed from dropdown as they are removed.

### DIFF
--- a/article/models.py
+++ b/article/models.py
@@ -26,6 +26,7 @@ from modelcluster.contrib.taggit import ClusterTaggableManager
 from section.sectionable.models import SectionablePage
 
 from taggit.models import TaggedItemBase
+from taggit.models import Tag
 
 from videos import blocks as video_blocks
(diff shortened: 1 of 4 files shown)


SamuelmdLow:
I fixed the initialization by slightly altering an inherited method to insert the value of the field into the choices list before they are rendered. This means the selected value will maintain being selected without having to load hundreds of tags just to delete them! But also this solution is incredible strange so not exactly happy about it.

PR #1686, opened 2025-12-01
progress bar in articles

### LINKED ISSUE #1677: Progress bar in articles

gm4vm0:

#1677 

1. added progress bar to `topbar` with flag `show_progress_bar=True`
2. removed underline from current section in header
3. added progress bar, same as 1


SamuelmdLow:
I think this could be improved if you moved the progress bar html into its own template within the navigation templates directory. That way you wouldn't have to duplicate html in multiple files. 

I think the progress bar would also look better inside some container element with a grey background colour. So the unfilled right-side part would be grey and would get filled in as blue as you scroll.

We can talk about it on Tuesday or you can message me with questions about the suggestions! Thanks for working on this issue!

Issue #1741, opened 2026-04-19
Article Creation: Reorder fields and add defaults


qsolace:
Some editors have expressed frustration with the ordering of some of the fields they need to enter when they are entering an article into wagtail. There is also a need to add some defaults to options that are almost always the exact same.
* Featured Image
  * There is always a featured image, so it doesn't make sense to make editors add a featured image each time
  * I think the change will be in one of the following:
https://github.com/ubyssey/ubyssey.ca/blob/dde202fcb06ba8d2ff03f76f870d69c27cdddb4d/article/models.py#L780-L787
https://github.com/ubyssey/ubyssey.ca/blob/dde202fcb06ba8d2ff03f76f870d69c27cdddb4d/article/models.py#L1604-L1612
* Article Content
  * This should initially be populated with a richtext block
  * See also:
 https://github.com/ubyssey/ubyssey.ca/blob/dde202fcb06ba8d2ff03f76f870d69c27cdddb4d/article/models.py#L1421-L1424
* Storystream view
  * This should be autopopulated with a default (apparently, it's automatically overriden when the senior masthead sets homepage)
 https://github.com/ubyssey/ubyssey.ca/blob/dde202fcb06ba8d2ff03f76f870d69c27cdddb4d/article/models.py#L674-L680
https://github.com/ubyssey/ubyssey.ca/blob/dde202fcb06ba8d2ff03f76f870d69c27cdddb4d/article/models.py#L858
https://github.com/ubyssey/ubyssey.ca/blob/dde202fcb06ba8d2ff03f76f870d69c27cdddb4d/article/models.py#L1694
* Authors page
  * This should be moved up to be between "Header/Banner Fields" and "Article Content"
  * I think this section should be moved up to 1612
https://github.com/ubyssey/ubyssey.ca/blob/dde202fcb06ba8d2ff03f76f870d69c27cdddb4d/article/models.py#L1624-L1631

- [x] Featured Image
- [x] Article Content
- [x] Storystream
- [x] Authors

PR #1746, opened 2026-04-21
1741 reorder article creation fields

qsolace:
Mostly just reorganizing a bunch of inputs for article creation — as described in #1741. I did go a little off script and hid the advanced fields in the liveblog, just because I think they're distracting and won't be used often.

I don't think there's much to look at here — I played around with all the page types and they seemed happy. I don't expect any major issues to come of this.


qsolace:
OH! Actually before we merge this I will show it to Spencer and get his sign off

qsolace:
checked with spencer and he approved the changes!

deacon8 (approved):

PR #1761, opened 2026-05-09
Remove Features, Humour, and Research from nav and homepage recent stories

jp-valino:
- Exclude features, humour, and research in get_recent_stories_by_day
- Exclude features, humour, and research in get_recent_stories_by_topic
- Complements the Wagtail menu and homepage section-row changes


qsolace:
#1762

qsolace (approved):
We may have to revert this in the future (if any of these sections return) but this is alright for now.

